PR #5, opened 2023-07-06
Add GH Action to publish to npm

JakeSCahill:
On merge to the `main` branch this Action will publish the package to npm only  if the version number in `package.json` differs from the latest on npm.

Uses https://github.com/marketplace/actions/npm-publish


JakeSCahill:
Checks failing because we need to merge #4

Deflaimun (approved):

PR #13, opened 2023-08-12
Remove images from Algolia index

JakeSCahill:
When we replaced the global `home` component with versioned home pages, this extension broke because we were pulling images from that component.

Removing images from the Algolia index for now. If needed, they can be added as a future improvement.


Feediver1 (approved):
lgtm

PR #23, opened 2023-12-04
Add data to schema

JakeSCahill:
Since we index a large number of records for docs now, this PR adds some best practices from Algolia by batching requests.

### DIFF
--- a/extensions/algolia-indexer/index.js
+++ b/extensions/algolia-indexer/index.js
@@ -3,9 +3,12 @@
 const generateIndex = require('./generate-index')
 const chalk = require('chalk')
 const algoliasearch = require('algoliasearch')
+const http = require('http')
+const https = require('https')
 const fs = require('fs')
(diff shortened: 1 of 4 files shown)


extensions/algolia-indexer/index.js
```diff
@@ -117,21 +119,11 @@ function register({
...
-          recordCount += batch.length
-        }
-      })
-
-      console.log(chalk.green('Total records:', recordCount))
-    } catch (err) {
-      logger.error(JSON.stringify(err))
-    }
+  process.on('exit', () => {
+    httpAgent.destroy()
```
Deflaimun, line 125: Isn't it better to change the flag from keepAlive to false? 

Does this get triggered both on success and error?

JakeSCahill: It should get triggered when the Node.js process is about to exit, regardless of the reason for the exit.

PR #28, opened 2024-01-18
Fetch latest RP Operator version and set in attribute

JakeSCahill:
On every docs build, fetches the latest release tag of Redpanda Operator and sets it as the value of the `latest-operator-version` attribute in the latest version of the `ROOT` component (Redpanda docs).

Part of https://github.com/redpanda-data/docs/pull/258


extensions/version-fetcher/get-latest-operator-version.js
```diff
@@ -0,0 +1,32 @@
...
+if (process.env.REDPANDA_GITHUB_TOKEN) {
+  githubOptions.auth = process.env.REDPANDA_GITHUB_TOKEN;
+}
+
+const github = new OctokitWithRetries(githubOptions);
+
+var latestOperatorReleaseVersion;
+
+module.exports = async () => {
+  await github.rest.repos.getLatestRelease({
```
chrisseto, line 22: It's a bit mind bending to me to see `await` intermixed with `.then`s. Is there a reason for that?
It also seems like `latestOperatorReleaseVersion` is meant to be a cache but it's not utilized as such?

Do you think something like this would work?
```js
// Used cached value, if set.
if (!!latestOperatorReleaseVersion) return latestOperatorReleaseVersion;

try {
    let release = await GitHub.rest.repos.getLatestRelease({owner, repo});
    latestOperatorReleaseVersion = release.data.tag_name;
    return latestOperatorReleaseVersion;
} catch (error) {
    console.error(error)
    return null;
}
```

chrisseto (approved):
LGTM!

extensions/version-fetcher/get-latest-operator-version.js
JakeSCahill: Thanks, Chris - updated to use `await` without the `then()`. Those values weren't cached. It runs once per build of the docs.

extensions/version-fetcher/get-latest-redpanda-version.js
```diff
@@ -0,0 +1,39 @@
...
+    const tag = release.data.tag_name;
+    latestRedpandaReleaseVersion = tag.replace('v', '');
+
+    // Get reference of the tag
+    const tagRef = await github.rest.git.getRef({ owner, repo, ref: `tags/${tag}` });
+    const releaseSha = tagRef.data.object.sha;
+
+    // Get the tag object to extract the commit hash
+    const tagData = await github.rest.git.getTag({ owner, repo, tag_sha: releaseSha });
+    latestRedpandaReleaseCommitHash = tagData.data.object.sha.substring(0, 7);
```
Deflaimun, line 32: We dont need the tag for the operator too?

Deflaimun (approved):

extensions/version-fetcher/get-latest-redpanda-version.js
JakeSCahill: This change gets the tag already. We don't need the extra tag data because we don't need the commit hash that's nested inside it. For Redpanda, we need the commit hash because it's returned by `rpk version`.

PR #39, opened 2024-05-03
Get latest Helm chart version

JakeSCahill:
As part of https://github.com/redpanda-data/documentation-private/issues/2403, this PR adds functionality to save the latest version of the Redpanda Helm chart to attributes in the latest docs for Redpanda.

It also fixes a bug where if the `latest` release tag on GitHub was set to an older patch release, the latest Redpanda version was set to that. Now, we sort the versions to make sure that we always return the actual latest version.

Note: I am seeing strange behavior in the GitHub API where 24.1.1 doesn't appear in https://api.github.com/repos/redpanda-data/redpanda/releases?per_page=1000

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@redpanda-data/docs-extensions-and-macros",
-  "version": "3.2.7",
+  "version": "3.2.8",
   "description": "Antora extensions and macros developed for Redpanda documentation.",
   "keywords": [
     "antora",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@redpanda-data/docs-extensions-and-macros",
-  "version": "3.2.8",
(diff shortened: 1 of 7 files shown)


Deflaimun (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@redpanda-data/docs-extensions-and-macros",
-  "version": "3.2.7",
+  "version": "3.2.9",
```
Deflaimun, line 3: shouldn't this be 8? xD

Deflaimun: Ah I see that https://github.com/redpanda-data/docs-extensions-and-macros/pull/40/files is 8

Deflaimun:
Should we remove the values from antora.yaml on main after this is merged? How do we deal with the conflict? Does this takes precedence? What about local builds where we might not have github token set?

PR #42, opened 2024-05-07
Return the latest version of Redpanda

JakeSCahill:
Since the GitHub API does not currently return version 24.1.1, we now support overwriting the version returned from GitHub with the one defined in the full-version attribute if that version is higher.


Deflaimun (approved):
good change. Using semver lib is a good idea too!

PR #66, opened 2024-08-21
Add Octokit core

JakeSCahill:
The npm docs for Octokit suggest this package is required. After some testing, I discovered that with this package, we no longer get errors when doing `npm upgrade`.


Deflaimun (approved):

Deflaimun:
check if the packages are incompatible with each other

Deflaimun (approved):

PR #69, opened 2024-08-27
Modify connect component visualization

Deflaimun:
Modify the visualization for connect components.  
Remove the datasource from the YAML and consider the CSV [redpanda_connect.csv](https://github.com/user-attachments/files/16874358/redpanda_connect.csv)

Remove the dropdown from the rows, and show a list of URLs instead.
Modify the filters to be multiple select. 
The filter works adding options. For example, 
- if a connector is of the type "input" and the filter have "input, output, processor" this connector will be shown.
- if a connector is of the type "output" and the filter have "input, output, processor" this connector will be shown.
- if a connector is of the type "processor" and the filter have "input, output " this connector will not be shown.

To run locally, run the rp-connect-docs repo, and modify the following to your local file 
https://github.com/redpanda-data/rp-connect-docs/blob/main/local-antora-playbook.yml#L28

For example: 'projects/antora/docs-extensions-and-macros/macros/rp-connect-components'

### DIFF
--- a/extensions/generate-rp-connect-info.js
+++ b/extensions/generate-rp-connect-info.js
@@ -55,7 +55,7 @@ module.exports.register = function ({ config,contentCatalog }) {
           for (const file of connectPages) {
             const filePath = file.path;
             if (filePath.endsWith(`${connector}.adoc`) && filePath.includes(`pages/${type}s/`)) {
-              url = `../${type}s/${connector}`;
+              url = `${type}s/${connector}`;
               break;
             }
           }
--- a/extensions/generate-rp-connect-info.js
+++ b/extensions/generate-rp-connect-info.js
@@ -9,74 +9,74 @@ const GITHUB_REF = 'connect-csv'
 /* const csvUrl = 'https://localhost:3000/csv';
 process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'; */
 
-module.exports.register = function ({ config,contentCatalog }) {
-    const logger = this.getLogger('redpanda-connect-info-extension');
+module.exports.register = function ({ config, contentCatalog }) {
+  const logger = this.getLogger('redpanda-connect-info-extension');
 
-    async function loadOctokit() {
-      const { Octokit } = await import('@octokit/rest');
-      if (!process.env.REDPANDA_GITHUB_TOKEN) return new Octokit()
-      return new Octokit({
-        auth: process.env.REDPANDA_GITHUB_TOKEN,
-      });
-    }
-
-    this.once('contentClassified', async ({ siteCatalog, contentCatalog }) => {
-        const redpandaConnect = contentCatalog.getComponents().find(component => component.name === 'redpanda-connect')
-        if (!redpandaConnect) return
-        const pages = contentCatalog.getPages()
-        try {
-            // Fetch CSV data and parse it
-            const csvData = await fetchCSV();
-            const parsedData = Papa.parse(csvData, { header: true, skipEmptyLines: true });
-            const enrichedData = enrichCsvDataWithUrls(parsedData,pages,logger);
-            parsedData.data = enrichedData
-            redpandaConnect.latest.asciidoc.attributes.csvData = parsedData;  
-            
-        } catch (error) {
-            logger.error('Error fetching or parsing CSV data:', error.message);
-            logger.error(error.stack);
-        }
+  async function loadOctokit() {
+    const { Octokit } = await import('@octokit/rest');
+    if (!process.env.REDPANDA_GITHUB_TOKEN) return new Octokit()
+    return new Octokit({
+      auth: process.env.REDPANDA_GITHUB_TOKEN,
     });
+  }
+
+  this.once('contentClassified', async ({ siteCatalog, contentCatalog }) => {
+    const redpandaConnect = contentCatalog.getComponents().find(component => component.name === 'redpanda-connect')
+    if (!redpandaConnect) return
+    const pages = contentCatalog.getPages()
+    try {
+      // Fetch CSV data and parse it
+      const csvData = await fetchCSV();
+      const parsedData = Papa.parse(csvData, { header: true, skipEmptyLines: true });
+      const enrichedData = enrichCsvDataWithUrls(parsedData, pages, logger);
+      parsedData.data = enrichedData
+      redpandaConnect.latest.asciidoc.attributes.csvData = parsedData;
+
+    } catch (error) {
+      logger.error('Error fetching or parsing CSV data:', error.message);
+      logger.error(error.stack);
+    }
+  });
 
   async function fetchCSV() {
-      const octokit = await loadOctokit();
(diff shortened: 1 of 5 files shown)


Deflaimun:
Animation123

Deflaimun:
I think the preview is not loading because netlify is not fetching the csv from github

JakeSCahill (commented):
This will work for the `component_type_dropdown` to use the CSV data:

```js
  registry.blockMacro(function () {
    const self = this;
    self.named('component_type_dropdown');
    self.process((parent, target, attrs) => {
      const attributes = parent.getDocument().getAttributes();
      const name = attributes['doctitle'];
      const type = attributes['type'];
  
      if (!name || !type) {
        return self.createBlock(parent, 'pass', '');
      }
  
      const csvData = context.config?.attributes?.csvData || [];
      const componentRows = csvData.data.filter(row => row.connector.trim().toLowerCase() === name.trim().toLowerCase());
  
      if (componentRows.length === 0) {
        return self.createBlock(parent, 'pass', '');
      }
  
      // Process types from CSV
      const types = componentRows.map(row => ({
        type: row.type.trim(),
        support: row.support_level.trim(),
        url: row.url ? row.url.trim() : '#'
      }));
  
      // Move the current page's type to the first position in the dropdown
      const sortedTypes = [...types];
      const currentTypeIndex = sortedTypes.findIndex(typeObj => typeObj.type === type);
      if (currentTypeIndex !== -1) {
        const [currentType] = sortedTypes.splice(currentTypeIndex, 1);
        sortedTypes.unshift(currentType);
      }
  
      // Check if the component requires an Enterprise license (based on support level)
      let enterpriseAdmonition = '';
      if (componentRows.some(row => row.support_level.toLowerCase() === 'enterprise')) {
        enterpriseAdmonition = `
        <div class="admonitionblock note">
          <table>
            <tbody>
              <tr>
                <td class="icon">
                  <i class="fa icon-note" title="Note"></i>
                </td>
                <td class="content">
                  <div class="paragraph">
                  <p>This feature requires an <a href="https://redpanda.com/compare-platform-editions" target="_blank">Enterprise license</a>. To upgrade, contact <a href="https://redpanda.com/try-redpanda?section=enterprise-trial" target="_blank" rel="noopener">Redpanda sales</a>.</p>
                  </div>
                </td>
              </tr>
            </tbody>
          </table>
        </div>`;
      }
  
      // Create the dropdown for types
      let typeDropdown = '';
      if (sortedTypes.length > 1) {
        typeDropdown = `
        <div class="page-type-dropdown">
          <p>Type: </p>
          <select class="type-dropdown" onchange="window.location.href=this.value">
            ${sortedTypes.map(typeObj => `<option value="../${typeObj.url}/" data-support="${typeObj.support}">${capitalize(typeObj.type)}</option>`).join('')}
          </select>
        </div>
        <script>
        // Initialize Choices.js for type dropdowns
        document.addEventListener('DOMContentLoaded', function() {
          const typeDropdowns = document.querySelectorAll('.type-dropdown');
          typeDropdowns.forEach(dropdown => {
            new Choices(dropdown, { searchEnabled: false, allowHTML: true, shouldSort: false });
          });
        });
        </script>`;
      }
      return self.createBlock(parent, 'pass', typeDropdown + enterpriseAdmonition);
    });
  });
```

Then we can remove everything apart from the code that builds the categories from `generate-rp-connect-categories.js`

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@redpanda-data/docs-extensions-and-macros",
-  "version": "3.5.11",
+  "version": "3.6.1",
```
JakeSCahill, line 3: We usually start from patch 0

macros/rp-connect-components.js
```diff
@@ -1,6 +1,117 @@
...
+      }
+      connectors[connector].supportLevels.get(support_level).add(commercial_name);
+      return connectors;
+    }, {});
+  }
+
+  function generateConnectorsHTMLTable(connectors, isCloud) {
+    return Object.entries(connectors).map(([connector, details], id) => {
+      const { types, supportLevels, isLicensed, urls } = details;
+      const firstUrl = urls.values().next().value || '#';
```
JakeSCahill, line 81: When we switch to Ash's CSV, they'll be a time when new connectors are in the catalog but we don't have docs for them yet. In this case, it's better to omit the link rather than provide an empty one to avoid users thinking that the link is broken.

macros/rp-connect-components.js
```diff
@@ -1,6 +1,117 @@
...
+          isLicensed: is_licensed === 'y' ? 'Yes' : 'No',
+          urls: new Set()
+        };
+      }
+      connectors[connector].types.set(capitalize(type), url);
+      if (url) connectors[connector].urls.add(url);
+      if (!connectors[connector].supportLevels.has(support_level)) {
+        connectors[connector].supportLevels.set(support_level, new Set());
+      }
+      connectors[connector].supportLevels.get(support_level).add(commercial_name);
```
JakeSCahill, line 73: I see you're relying on there being multiple rows for a `connector` and the `commercial_name` being the specific flavor (driver, cache etc..). But Ash's CSV does not follow this pattern. We'll need to update this logic to use whatever Ash has here: https://github.com/redpanda-data/connect/pull/2827#discussion_r1743287405

For example, the support for each `cache` output target should be fetched from the `cache` connector types.

`cache` connectors:

```csv
cache                   ,output    ,cache                   ,certified  ,n          ,y     ,y
cache                   ,processor ,cache                   ,certified  ,n          ,y     ,y
```

The support levels for each cache type is defined in the connectors with type `cache`:

```
gcp_cloud_storage       ,cache     ,gcp_cloud_storage       ,community  ,n          ,n     ,n
lru                     ,cache     ,lru                     ,community  ,n          ,y     ,y
```

extensions/generate-rp-connect-info.js
```diff
@@ -0,0 +1,71 @@
+'use strict'
+const https = require('https');
+const Papa = require('papaparse');
+
+const csvUrl = 'https://raw.githubusercontent.com/redpanda-data/rp-connect-docs/connect-csv/redpanda_connect.csv';
```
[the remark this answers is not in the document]
JakeSCahill: We may need to set `GITHUB_REF` to the latest tag for RP Connect. See https://redpandadata.slack.com/archives/C07KNARFF9C/p1725440899440089

In this case, we can reuse the module in `extensions/version-fetcher/get-latest-connect.js`

JakeSCahill:
> I think the preview is not loading because netlify is not fetching the csv from github

it's because `redpanda-connect` docs aren't available in this build. You can add them to the local playbook:

```yml
- url: https://github.com/redpanda-data/rp-connect-docs
    branches: main
```

then you can also see the catalog preview from this build.

package.json
Deflaimun, line 3:
```suggestion
  "version": "3.6.0",
```

PR #76, opened 2024-09-11
Set anchors to term titles

JakeSCahill:
This PR addresses an issue where the `term-name` attribute was being used to generate anchor links for glossary terms. In some cases, the actual term title displayed on the page differed from the `term-name`, leading to broken or mismatched links.

To ensure consistency between the displayed term title and the anchor link, this update modifies the macro to always use the actual term title when building anchor links. This change guarantees that links point to the correct location based on the title shown on the page.

### DIFF
--- a/macros/glossary.js
+++ b/macros/glossary.js
@@ -27,6 +27,11 @@ module.exports.register = function (registry, config = {}) {
 
       const terms = termFiles.map(file => {
         const content = file.contents.toString()
+        // Split content by lines and get the first non-empty line as the title
+        const lines = content.split('\n').map(line => line.trim())
+        const firstNonEmptyLine = lines.find(line => line.length > 0)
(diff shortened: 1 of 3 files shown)


macros/glossary.js
```diff
@@ -76,13 +82,16 @@ module.exports.register = function (registry, config = {}) {
     }
   }
 
-  //characters to replace by '-' in generated idprefix
-  const IDRX = /[/ _.-]+/g
+  // Characters to replace by '-' in generated idprefix (no brackets in this regex)
```
JakeSCahill, line 85:
```suggestion
  // Characters to replace by '-' in generated idprefix
```

Deflaimun (approved):
lgtm. Notice that if the page doesn't have the necessary height, it can't jump directly into the term. Altough I believe this is intended/expected.

PR #77, opened 2024-09-12
Add support for local files

JakeSCahill:
This PR adds support for providing local terms files as well as a local attributes file. This feature is useful for local development where you want to test the extension or you don't want to rely on loading the `shared` component from GitHub or the local filesystem.

For details, see https://github.com/redpanda-data/docs-extensions-and-macros/pull/77/files#diff-1194561cbea346acb1bea37d81fdf5c70def982dcbb0c64eca2a96bc51941124

This PR also adds better log messages and fixes a bug where you couldn't build the docs without the `shared` component available in the playbook.

### DIFF
--- a/extensions/add-global-attributes.js
+++ b/extensions/add-global-attributes.js
@@ -4,33 +4,83 @@
  *    - require: ./extensions/add-global-attributes.js
 */
 
+const fs = require('fs');
+const path = require('path');
+const yaml = require('js-yaml');
+const _ = require('lodash');
+
+const ATTRIBUTES_PATH = 'modules/ROOT/partials/';  // Default path within the 'shared' component
+
 module.exports.register = function ({ config }) {
-  const yaml = require('js-yaml');
   const logger = this.getLogger('global-attributes-extension');
-  const chalk = require('chalk')
-  const _ = require('lodash');
+  const chalk = require('chalk');
+
+  /**
+   * Load global attributes from a specified local file if provided.
+   */
+  function loadLocalAttributes(siteCatalog, localAttributesFile) {
+    try {
+      const resolvedPath = path.resolve(localAttributesFile);
+      if (!fs.existsSync(resolvedPath)) {
+        logger.warn(`Local attributes file "${localAttributesFile}" does not exist.`);
+        return false;  // Return false if the local file doesn't exist
+      }
+
+      const fileContents = fs.readFileSync(resolvedPath, 'utf8');
+      const fileAttributes = yaml.load(fileContents);
+
+      siteCatalog.attributeFile = _.merge({}, fileAttributes);
+      console.log(chalk.green(`Loaded global attributes from local file "${localAttributesFile}".`));
+      return true;  // Return true if the local attributes were successfully loaded
+
+    } catch (error) {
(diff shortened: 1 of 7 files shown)


extensions/add-global-attributes.js
```diff
@@ -4,33 +4,83 @@
...
+  const chalk = require('chalk');
+
+  /**
+   * Load global attributes from a specified local file if provided.
+   */
+  function loadLocalAttributes(siteCatalog, localAttributesFile) {
+    try {
+      const resolvedPath = path.resolve(localAttributesFile);
+      if (!fs.existsSync(resolvedPath)) {
+        logger.warn(`Local attributes file "${localAttributesFile}" does not exist.`);
```
Deflaimun, line 25: not for now, but we might consider adding a new npm goal with verbose logging.

This one for example, could be omitted in most cases.

extensions/add-global-attributes.js
```diff
@@ -4,33 +4,83 @@
...
+      }
+
+      // If no local attributes were loaded, fallback to the 'shared' component
+      if (!attributesLoaded) {
+        let sharedComponentFound = false;
+
+        for (const component of contentAggregate) {
+          if (component.name === 'shared') {
+            sharedComponentFound = true;
+            const attributeFiles = component.files.filter(file => file.path.startsWith(ATTRIBUTES_PATH) && file.path.endsWith('.yml'));
```
Deflaimun, line 58: Looks like both yml and yaml are supported suffixes 
https://stackoverflow.com/questions/21059124/is-it-yaml-or-yml

Deflaimun, line 58:
```suggestion
            const attributeFiles = component.files.filter(file => file.path.startsWith(ATTRIBUTES_PATH) && (file.path.endsWith('.yml') || file.path.endsWith('.yaml'));
```

extensions/aggregate-terms.js
```diff
@@ -27,32 +35,91 @@ module.exports.register = function ({ config }) {
...
+        return false;
+      }
+
+      termFiles.forEach(file => {
+        const filePath = path.join(resolvedPath, file);
+        const termContent = fs.readFileSync(filePath, 'utf8');
+        const categoryMatch = /:category: (.*)/.exec(termContent);
+        const category = categoryMatch ? categoryMatch[1] : 'Miscellaneous';
+
+        const formattedCategory = category.charAt(0).toUpperCase() + category.slice(1);
```
Deflaimun, line 61: We could consider adding this capitalize function in an util file 
https://github.com/redpanda-data/docs-extensions-and-macros/blob/main/macros/rp-connect-components.js#L49
https://github.com/redpanda-data/docs-extensions-and-macros/tree/main/extensions/util

Deflaimun (approved):
lgtm. please see the comment about logging before merging

extensions/add-global-attributes.js
JakeSCahill: I used `warn` here because it fits the descriptions of the Antora log levels: https://docs.antora.org/antora/latest/playbook/runtime-log-level/

We could have a cut-down playbook that has `fatal` logs on only. Or we could document how to do that in a wiki so anyone can customize the local playbook as necessary.

We can discuss more after next week 😄

PR #78, opened 2024-09-18
DOC-549 Add support for a master component table and other metadata on connector docs

JakeSCahill:
Fixes https://github.com/redpanda-data/documentation-private/issues/2733

This PR improves visibility into whether connectors are supported in Redpanda Cloud, along with a unified master table displaying key metadata for all connectors. It also **updates the automation to pull the CSV data from the source of truth at https://github.com/redpanda-data/connect/blob/main/internal/plugins/info.csv** instead of our local copy in the docs, which isn't kept up to date.

On each connector doc, we indicate in which products it is supported. When both Cloud and Self-Managed versions are available, users can easily switch between them via clearly labeled links. The current version is indicated using CSS without the need for additional text, reducing noise on the page. CSS styles are added in https://github.com/redpanda-data/docs-ui/pull/223

## Previews

[Master table](https://deploy-preview-78--docs-extensions-and-macros.netlify.app/preview/test/#connector-table-with-all-data)
[Metadata](https://deploy-preview-78--docs-extensions-and-macros.netlify.app/redpanda-cloud/develop/connect/components/inputs/amqp_0_9/)

NOTE: The metadata will look like this when the following PR merged: https://github.com/redpanda-data/docs-ui/pull/223

### DIFF
--- a/extensions/generate-rp-connect-info.js
+++ b/extensions/generate-rp-connect-info.js
@@ -28,6 +28,7 @@ module.exports.register = function ({ config }) {
   this.once('contentClassified', async ({ contentCatalog }) => {
     const redpandaConnect = contentCatalog.getComponents().find(component => component.name === 'redpanda-connect');
     const redpandaCloud = contentCatalog.getComponents().find(component => component.name === 'redpanda-cloud');
+    const preview = contentCatalog.getComponents().find(component => component.name === 'preview');
     if (!redpandaConnect) return;
     const pages = contentCatalog.getPages();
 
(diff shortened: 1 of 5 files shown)


vladoschreiner:
LGTM

Minor suggestions to the master table
1. "Available in cloud" instead of cloud support. Cloud support naming suggests that the Support level column applies to cloud, too
2. "Available in cloud" filter, too

vladoschreiner (approved):

Deflaimun (approved):

PR #80, opened 2024-10-30
Support beta in tests

JakeSCahill:
To support testing Docker Compose files for both beta (unstable) and GA (general availability) versions, this PR modifies the  extension to dynamically replace specific environment variables based on the release type. These environment variables are used in our automated tests, but we don't want to force users to set them, so this change replaces them dynamically.

- If `page-component-version-is-prerelease` is set, `${REDPANDA_DOCKER_REPO:-redpanda}` is replaced with `redpanda-unstable`, indicating the beta Docker repo.
- If `page-component-version-is-prerelease` is not set, the placeholder defaults to `redpanda` for the GA Docker repo.

Here's an example snippet from the quickstart:
```yaml
    image: docker.redpanda.com/redpandadata/${REDPANDA_DOCKER_REPO:-redpanda}:${REDPANDA_VERSION:-latest}
    container_name: redpanda-0
```
In this case, our Doc Detective tests will set the environment variables or use the fallbacks so we can test during GA and beta. But we want to single-source the same file into docs without requiring users to set them. So, the extension is responsible for finding `REDPANDA_DOCKER_REPO` and dynamically changing it to `redpanda-unstable` when the doc is a beta or `redpanda` otherwise.

We also now capture raw tags to avoid having to add the `v` prefix for certain use cases, which is error-prone. So each `latest-<product>` attribute has both a `version` and a `tag` option. `version` is the raw version number and `tag` is the release tag, including any `v` prefixes.

### DIFF
--- a/extensions/version-fetcher/set-latest-version.js
+++ b/extensions/version-fetcher/set-latest-version.js
@@ -77,11 +77,8 @@ module.exports.register = function ({ config }) {
 
           // Special handling for Redpanda RC versions if in beta
           if (latestVersions.redpanda?.latestRcRelease?.version) {
-            const betaVersion = sanitizeVersion(latestVersions.redpanda.latestRcRelease.version);
-            asciidoc.attributes['redpanda-beta-version'] = betaVersion;
-            asciidoc.attributes['redpanda-beta-tag'] = `v${betaVersion}`;
+            setVersionAndTagAttributes(asciidoc, 'redpanda-beta', latestVersions.redpanda.latestRcRelease.version, name, version)
             asciidoc.attributes['redpanda-beta-commit'] = latestVersions.redpanda.latestRcRelease.commitHash;
-            logger.info(`Set Redpanda RC version ${betaVersion} (tag: v${betaVersion}) in ${name} ${version}`);
           }
         });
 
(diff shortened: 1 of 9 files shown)


extensions/version-fetcher/set-latest-version.js
```diff
@@ -17,83 +20,108 @@ module.exports.register = function ({ config }) {
...
-          if (prerelease && prerelease.version === version) {
-            asciidoc.attributes['page-component-version-is-prerelease'] = 'true'
+          if (prerelease?.version === version) {
+            asciidoc.attributes['page-component-version-is-prerelease'] = 'true';
           }
-          if (LatestConsoleVersion) {
-            asciidoc.attributes['latest-console-version'] = `${LatestConsoleVersion}@`
-            logger.info(`Set Redpanda Console version to ${LatestConsoleVersion} in ${name} ${version}`)
+
+          // Conditionally set operator and helm chart attributes only if data is available
```
JakeSCahill, line 62:
```suggestion
          // Set operator and helm chart attributes
```

extensions/version-fetcher/set-latest-version.js
```diff
@@ -17,83 +20,108 @@ module.exports.register = function ({ config }) {
...
-}
\ No newline at end of file
+  });
+
+  // Helper function to set both latest-*version and latest-*tag attributes
+  function setVersionAndTagAttributes(asciidoc, baseName, versionData, name = '', version = '') {
+    if (versionData) {
+      const versionWithoutPrefix = sanitizeVersion(versionData);
+      asciidoc.attributes[`${baseName}-version`] = versionWithoutPrefix; // Without "v" prefix
+      asciidoc.attributes[`${baseName}-tag`] = `v${versionWithoutPrefix}`; // With "v" prefix
```
JakeSCahill, line 113:
```suggestion
      asciidoc.attributes[`${baseName}-tag`] = `${versionData}`;
```

Deflaimun:
Does this change affect other pages that are not the quickstart?

JakeSCahill:
> Does this change affect other pages that are not the quickstart?

It'll affect all YAML files in Self-Managed 24.3 (including beta) onwards.

Deflaimun (approved):

PR #112, opened 2025-07-16
Allow for unknown types

JakeSCahill:
`unknown` is a valid type in Redpanda Connect (not a bug like we previously thought). This PR removes the logic that maps `unknown` types to objects: https://github.com/redpanda-data/connect/blob/main/docs/modules/configuration/pages/templating.adoc#fieldstype


mihaitodor (approved):
:shipit:

PR #118, opened 2025-07-21
New metrics automation 

paulohtb6:
# New metrics automation

Introducing a new metrics automation tool that traverses through the Repdanda source code to create reliable automation.

Run it: 
```
npx generate metrics-docs --tag [tag-value. Can be dev or v25.2.4-rc4 for example]
```

## Changes

Moved the old metrics command to `metric-docs-legacy` 

## Json output schema
```json
{
  "redpanda_kafka_requests_total": {
    "unique_id": "a1b2c3d4e5f6...",
    "name": "redpanda_kafka_requests_total",
    "full_name": "redpanda_kafka_requests_total",
    "metric_type": "external",
    "type": "counter",
    "description": "Total number of Kafka requests.",
    "labels": ["request_type", "status"],
    "constructor": "make_counter",
    "files": [{"file": "src/v/kafka/server/handlers/handler.cc", "line": 45}],
    "group": "kafka"
  }
}
```

## more info
More info at the provided readmes


JakeSCahill:
@coderabbitat review

tools/metrics/metrics.py
```diff
@@ -120,6 +120,27 @@ def parse_metrics(metrics_text):
     logging.info(f"Extracted {len(metrics)} metrics.")
     return metrics
 
+def filter_metrics_for_docs(metrics):
+    """Filter metrics for documentation - remove duplicates and histogram suffixes."""
+    filtered = {}
+    seen_names = set()  # Track metric names to detect duplicates
+    
+    for name, data in metrics.items():
```
JakeSCahill, line 128: Nice!

bin/doc-tools.js
```diff
@@ -713,6 +792,41 @@ automation
     process.exit(0);
   });
 
+automation
+  .command('source-metrics-docs')
```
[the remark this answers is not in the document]
JakeSCahill: Please also keep support for specifying paths for the output

bin/doc-tools.js
```diff
@@ -429,7 +460,51 @@ function diffDirs(kind, oldTag, newTag) {
...
+    console.log(`✅ Enhanced metrics extraction completed! (${duration}s)`);
+    console.log(`📄 Generated files:`);
+    console.log(`   JSON: ${options.jsonOutput}`);
+    console.log(`   Internal metrics: ${options.internalAsciidoc}`);
+    console.log(`   Public metrics: ${options.externalAsciidoc}`);
+    process.exit(0);
+  });
+
+automation
+  .command('metrics-docs-legacy')
```
JakeSCahill, line 506: I would just remove the legacy implementation.

package.json
```diff
@@ -21,7 +21,9 @@
     "get-console-version": "doc-tools get-console-version",
     "build": "antora --to-dir docs --fetch local-antora-playbook.yml",
     "serve": "wds --node-resolve --open preview/test/ --watch --root-dir docs",
-    "test": "jest"
+    "test": "jest",
```
JakeSCahill, line 24: Please bump the version in this file. Since this is quite a large change, bump the minor version to 4.8.0 and then `npm install` to update the lock file.

JakeSCahill:
Looking really nice @paulohtb6 thanks for working on this!

When it's ready, can you update our wiki: https://redpandadata.atlassian.net/wiki/spaces/DOC/pages/1185054748/Doc+Tools+CLI

paulohtb6:
During testing I identified several false-positives from the generation. Most of the problems were because the parsing was failing to get group_names. Some are EXTREMELY hard to identify via this method, even using the AST. 

One in particular I'm considering impossible, because it spawns multiple files. 

I'm trying to reduce the number of false-positives, and the latest commit introduces an aggressive check to resolve the group_names.

paulohtb6:
Just noticed that some labels are also missclassified. 

In these two metrics, the autogenerated labels look like this. 

```adoc
=== redpanda_trust_file_crc32c

CRC32C checksum calculated from the contents of the trust file. This value is calculated when a valid certificate is loaded and a trust store is present. Otherwise, the value is zero.

*Type*: gauge

*Labels*:

- `aggregates`
- `label`
- `shard`

---

=== redpanda_truststore_expires_at_timestamp_seconds

Expiry time of the shortest-lived CA in the truststore(seconds since epoch).

*Type*: gauge

*Labels*:

- `aggregates`
- `label`
- `label_instance`

---
```

On the source code, available here https://github.com/redpanda-data/redpanda/blob/dev/src/v/net/probes.cc we can tell that it's actually [`area`](https://github.com/redpanda-data/redpanda/blob/dev/src/v/net/probes.cc#L347), [`detail`](https://github.com/redpanda-data/redpanda/blob/dev/src/v/net/probes.cc#L348), [`shard` ](https://github.com/redpanda-data/redpanda/blob/dev/src/v/net/probes.cc#L418)

PR #123, opened 2025-08-05
Properties aliases

paulohtb6:
Adds the ability to write property aliases into the generated pages.


JakeSCahill (approved):
Nice addition. TY

PR #158, opened 2025-12-03
gha: update publish-to-npm (using OIDC now)

clee:
Should fix [DEVPROD-3689], as I've already set up the OIDC link on the NPMJS side.

[DEVPROD-3689]: https://redpandadata.atlassian.net/browse/DEVPROD-3689


JakeSCahill (approved):
Thanks!

PR #179, opened 2026-03-27
Wrap `$ref` with sibling properties in `allOf` during OpenAPI spec bundling

kbatuigas:
Related: https://github.com/redpanda-data/api-docs/pull/60

In OpenAPI 3.0, properties placed alongside $ref (like description and readOnly) are [silently ignored](https://www.learnjsonschema.com/draft4/core/ref/). This causes our rendered API docs on Bump.sh to show the referenced schema's description instead of the more specific inline description.

This PR adds a post-processing step to `bundle-openapi` that wraps $ref nodes having sibling properties in `allOf`:

```
# Before                          # After
passwordSetAt:                    passwordSetAt:
  $ref: "...Timestamp"              allOf:
  description: "The timestamp…"        - $ref: "...Timestamp"
  readOnly: true                    description: "The timestamp…"
                                    readOnly: true
```

The result is already implemented in https://github.com/redpanda-data/api-docs/blob/main/admin/admin-v2.yaml (For example this [instance](https://github.com/redpanda-data/api-docs/blob/main/admin/admin-v2.yaml#L1738) of `passwordsetAt` displays [this description](https://docs.redpanda.com/api/doc/admin/v2/operation/operation-redpanda-core-admin-v2-securityservice-getscramcredential#operation-redpanda-core-admin-v2-securityservice-getscramcredential-200-body-application-json-scramcredential-passwordsetat) in the docs, and not `google.protobuf.Timestamp`'s)

This is the standard workaround for OpenAPI 3.0 and has been confirmed as the correct approach by Bump.sh's developer team. The transformation is done during bundling and no changes are needed to source specs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ gen/
 tree-sitter/
 test/
 test-results/
+coverage/
 __tests__/.test-cgo-detection/
 modules/
 cloud-controlplane/
--- a/__tests__/tools/bundle-openapi.test.js
+++ b/__tests__/tools/bundle-openapi.test.js
@@ -11,6 +11,7 @@ const {
   normalizeTag,
   getMajorMinor,
   sortObjectKeys,
+  wrapRefSiblings,
   detectBundler,
   createEntrypoint,
   runBundler,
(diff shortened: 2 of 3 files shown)


JakeSCahill (approved):
LGTM

PR #189, opened 2026-04-10
fix: ensure llms directive HTML matches afdocs test pattern

JakeSCahill:
## Summary

Updates the add-llms-directive extension to ensure generated HTML matches the afdocs test pattern expectations. The afdocs test uses a regex pattern that expects specific formatting in the HTML directive.

## Changes

- Added explicit documentation comment about afdocs pattern requirements
- Bumped version to 4.16.2

## Context

The afdocs test was reporting "llms.txt directive found in 1 of 49 sampled pages (48 missing)" even though manual verification shows the directive is present on all pages. This PR ensures the generated HTML explicitly matches the expected pattern.


micheleRP (approved):
lgtm

PR #191, opened 2026-04-17
fix: correct three bugs in Redpanda Connect connector documentation generation

JakeSCahill:
## Summary

This PR fixes three critical bugs in the Redpanda Connect connector documentation generator that were discovered in PR https://github.com/redpanda-data/rp-connect-docs/pull/414:

### Bug 1: "metricss" double 's' in paths

**Problem**: Include paths for the `open_telemetry_collector` metrics component were generated as `metricss/` instead of `metrics/`:
```
include::components:example$common/metricss/open_telemetry_collector.yaml[]
```

**Root cause**: The code in `rpcn-connector-docs-handler.js` (lines 1132, 1149) was singularizing all types by removing the trailing 's', converting "metrics" → "metric". Then the Handlebars template was blindly appending 's' to create the plural form, resulting in "metricss".

**Fix**: Added special handling for "metrics" type which is both singular and plural. Updated the template to conditionally pluralize based on the component type.

**Files changed**:
- `tools/redpanda-connect/rpcn-connector-docs-handler.js`: Lines 1132, 1149
- `tools/redpanda-connect/templates/connector.hbs`: Lines 20, 29, 50, 54

---

### Bug 2: Arc description truncation with malformed output

**Problem**: The `arc` output description in `whats-new.adoc` was truncated and malformed:
```
|com/Basekick-Labs/arc[Arc^] columnar analytical database...
```

Should have been:
```
|This output sends data to an https://github.com/Basekick-Labs/arc[Arc^] columnar analytical database...
```

**Root cause**: The `capToTwoSentences()` function in `rpcn-connector-docs-handler.js` was treating the period in `https://github.com` as a sentence terminator. The sentence detection regex would find the period after "github" and start capturing the next "sentence" from "com/", losing the beginning of the description.

**Fix**: Added URL pattern (`/https?:\/\/[^\s]+/gi`) to the abbreviations array (line 30) so that URLs are protected from sentence detection by replacing them with placeholders before processing.

**Files changed**:
- `tools/redpanda-connect/rpcn-connector-docs-handler.js`: Line 30

---

### Bug 3: CGO components appearing as "new" every release

**Problem**: CGO-only components like `tigerbeetle_cdc`, `zmq4`, and `ffi` were incorrectly appearing as "new connectors" in every release, even though they existed in the previous version (4.87.0).

**Root cause**: Lines 1370-1377 in `rpcn-connector-docs-handler.js` had faulty logic. When `oldBinaryAnalysis` was unavailable, the code would check if CGO components existed in the old OSS index:
```javascript
const wasInOldOss = oldIndex[cgoComp.type]?.some(c => c.name === cgoComp.name)
return !wasInOldOss  // Returns TRUE for CGO-only components!
```

Since CGO-only components are NEVER in the OSS index, this always returned `true`, treating them as new.

**Fix**: Updated the logic to check if the component existed in the old index with `requiresCgo: true` flag, since CGO components are saved in the connector data files with this flag.

**Files changed**:
- `tools/redpanda-connect/rpcn-connector-docs-handler.js`: Lines 1370-1377

---

## Testing

All three bugs have been verified as fixed in PR https://github.com/redpanda-data/rp-connect-docs/pull/414:

1. ✅ `open_telemetry_collector` now generates correct paths: `metrics/open_telemetry_collector.yaml`
2. ✅ `arc` description is complete: "This output sends data to an https://github.com/Basekick-Labs/arc[Arc^]..."
3. ✅ CGO components (`tigerbeetle_cdc`, `zmq4`, `ffi`) no longer appear as new

## Related PRs

- https://github.com/redpanda-data/rp-connect-docs/pull/414 - PR that exposed these bugs


micheleRP:
## CodeRabbit findings worth considering

Surfacing three CodeRabbit items from its automated review — two are worth acting on before merge, the third is a nice-to-have cleanup.

### 1. URL-abbreviation regex is too greedy (`tools/redpanda-connect/rpcn-connector-docs-handler.js:31`)

The bug 2 fix uses `/https?:\/\/[^\s]+/gi` to protect URLs. That pattern also swallows trailing sentence terminators. Example: `"See https://a.com/x. Second sentence."` → match is `https://a.com/x.` (period consumed) → sentence boundary lost, `maxSentences = 2` misbehaves. Fix for Arc works, but may introduce a new class of truncation.

Suggested tighter regex:

```js
/https?:\/\/[^\s]+?(?=[.!?](?:\s|$)|\s|$)/gi
```

Worth adding a test for a URL that ends a sentence.

### 2. `requiresCgo === undefined` fallback (`tools/redpanda-connect/rpcn-connector-docs-handler.js:1376-1391`)

When `oldBinaryAnalysis` is unavailable, the fallback checks `c.requiresCgo === true || c.requiresCgo === undefined` against an index whose `requiresCgo` is explicitly stripped by `stripPlatformMetadata()`. The `=== undefined` branch degenerates into a bare name lookup — which is exactly the original-bug behavior this PR claims to fix. The subsequent `!docsExist` check papers over it, but the logic is circular.

Either drop the `=== undefined` branch, or add a comment making the "treat missing metadata as possibly CGO" intent explicit.

### 3. Handlebars template duplication (`tools/redpanda-connect/templates/connector.hbs:20-54`) — nitpick

The `{{#if (eq type 'metrics')}}{{type}}{{else}}{{type}}s{{/if}}` expression repeats four times (lines 20, 29, 50, 54). If another "keep-as-plural" type is added later, all four call sites must be kept in sync. Options:

- Register a Handlebars helper `typeDir` that returns `'metrics'` for metrics and `${type}s` otherwise — call sites become `{{typeDir type}}`, or
- Compute a `typeDir` property on the render context in `generate-rpcn-connector-docs.js` before invoking the template — call sites become `{{typeDir}}`.

Either centralizes the pluralization rule next to the singularization logic in `rpcn-connector-docs-handler.js` (lines 1132, 1151). Not blocking — can be a follow-up.

micheleRP (approved):
please see Claude's suggestions before merge

PR #194, opened 2026-04-27
fix: skip field-only pages in unlisted pages check

JakeSCahill:
## Summary
- Skip pages marked with `isFieldOnlyPage = true` in the unlisted-pages extension
- Eliminates noisy warnings for `/tmp/generated-fields-only/` paths during Antora builds
- These pages are intentionally not in navigation (they're internal includes, not standalone content)

## Test plan
- [x] All 370 tests pass
- [x] Build rp-connect-docs and verify no warnings for field-only pages


micheleRP (approved):
lgtm

PR #199, opened 2026-05-27
Fix old component name in attribute replacer extension

JakeSCahill:
This pull request includes a version bump for the `@redpanda-data/docs-extensions-and-macros` package and an update to the logic for determining when to use tag attributes in `replace-attributes-in-attachments.js`.

Version update:

* Bumped the package version in `package.json` from `5.0.1` to `5.0.2`.

Logic update:

* Changed the condition in `getDynamicReplacements` so that the `is24_3plus` flag now checks for `componentVersion.title === 'Streaming'` instead of `'Self-Managed'`. This affects how tag attributes are applied for different component versions.

### DIFF
--- a/extensions/replace-attributes-in-attachments.js
+++ b/extensions/replace-attributes-in-attachments.js
@@ -181,7 +181,7 @@ function getDynamicReplacements(componentVersion, logger) {
   const isPrerelease = attrs['page-component-version-is-prerelease'];
   const versionNum = formatVersion(componentVersion.version || '', semver);
   const is24_3plus =
-    versionNum && semver.gte(versionNum, '24.3.0') && componentVersion.title === 'Self-Managed';
+    versionNum && semver.gte(versionNum, '24.3.0') && componentVersion.title === 'Streaming';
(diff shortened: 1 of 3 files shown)


micheleRP (approved):
LGTM — verified across `v/24.1` through `main` that the component `title:` is now `Streaming` everywhere, so the old `'Self-Managed'` check could never fire after the rename. Fix restores the intended `useTagAttributes` behavior for 24.3+ attachments.

### Suggestions (non-blocking)

1. **`extensions/replace-attributes-in-attachments.js:184-185`** — Both `'Streaming'` and `'Labs'` are magic strings tied to upstream `antora.yml` `title:` values. A small constants block at the top of the file (e.g., `const STREAMING_TITLE = 'Streaming'; const LABS_TITLE = 'Labs';`) would make the next rename easier to grep for.

2. **No test coverage for `getDynamicReplacements`.** `__tests__/extensions/` has no test for this file, which is why the regression went unnoticed when the component was renamed. Worth a follow-up — out of scope here.

PR #200, opened 2026-06-01
Add available-properties-tag and available-connect-version attributes

JakeSCahill:
## Summary

The `set-latest-version` extension now scans the content catalog for available JSON attachment files and sets attributes that point to versions that actually exist:

- `available-properties-tag`: Latest `redpanda-properties-*.json` version found in `streaming:reference` module
- `available-connect-version`: Latest `connect-*.json` version found in `connect:components` module

## Problem

The existing `latest-redpanda-tag` and `latest-connect-version` attributes are set from GitHub releases, which may point to versions that don't have docs/attachments published yet. This causes unresolved resource errors when the UI tries to load JSON files for property tooltips.

## Solution

After fetching GitHub release versions, the extension now:
1. Scans the content catalog for attachment files matching the JSON patterns
2. Extracts versions from filenames and sorts by semver
3. Sets `available-*` attributes to the highest available version

The UI can then use `available-properties-tag` (with fallback to `latest-redpanda-tag`) to ensure it always references a JSON file that exists.

## Changes

- **`extensions/version-fetcher/set-latest-version.js`**: Added content catalog scanning logic
- **`__tests__/extensions/version-fetcher/available-json-versions.test.js`**: 25 unit tests

## Backwards Compatibility

- Supports both `streaming` and `ROOT` component names
- Supports both `connect` and `redpanda-connect` component names
- New attributes are additive - existing attributes unchanged

## Test plan

- [x] All 25 unit tests pass
- [ ] Test with local Antora build to verify attributes are set correctly

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/__tests__/extensions/version-fetcher/available-json-versions.test.js
+++ b/__tests__/extensions/version-fetcher/available-json-versions.test.js
@@ -33,7 +33,11 @@ function filterConnectFiles(files) {
 function extractLatestPropertiesVersion(files) {
   const versions = files
     .map(f => f.src.stem.replace('redpanda-properties-', ''))
-    .filter(v => semver.valid(v.replace(/^v/, '')));
+    // Only accept stable versions (no prerelease suffix like -backup, -rc1, etc.)
+    .filter(v => {
+      const normalized = v.replace(/^v/, '');
+      return semver.valid(normalized) && semver.prerelease(normalized) === null;
+    });
 
   if (versions.length === 0) return null;
 
@@ -45,7 +49,8 @@ function extractLatestPropertiesVersion(files) {
 function extractLatestConnectVersion(files) {
   const versions = files
     .map(f => f.src.stem.replace('connect-', ''))
-    .filter(v => semver.valid(v));
+    // Only accept stable versions (no prerelease suffix)
+    .filter(v => semver.valid(v) && semver.prerelease(v) === null);
 
   if (versions.length === 0) return null;
 
@@ -170,30 +175,31 @@ describe('Properties JSON file detection', () => {
       expect(result).toBeNull();
     });
 
-    it('should handle prerelease versions correctly', () => {
+    it('should reject prerelease versions and use stable', () => {
       const files = [
         { src: { stem: 'redpanda-properties-v25.3.1' } },
         { src: { stem: 'redpanda-properties-v26.0.0-rc1' } },
       ];
 
       const result = extractLatestPropertiesVersion(files);
-      // v26.0.0-rc1 is higher than v25.3.1 in semver (26 > 25)
-      // If a prerelease JSON exists, we should use it
-      expect(result).toBe('v26.0.0-rc1');
+      // v26.0.0-rc1 is filtered out as a prerelease, so v25.3.1 wins
+      expect(result).toBe('v25.3.1');
     });
 
-    it('should prefer stable over prerelease of same major.minor', () => {
+    it('should reject all prerelease suffixes (-rc1, -beta1, -backup, etc.)', () => {
       const files = [
         { src: { stem: 'redpanda-properties-v25.3.1' } },
         { src: { stem: 'redpanda-properties-v25.3.2-rc1' } },
+        { src: { stem: 'redpanda-properties-v26.0.0-backup' } },
+        { src: { stem: 'redpanda-properties-v25.4.0-beta1' } },
       ];
 
       const result = extractLatestPropertiesVersion(files);
-      // v25.3.2-rc1 < v25.3.2 but > v25.3.1, so rc1 wins
-      expect(result).toBe('v25.3.2-rc1');
+      // All prerelease/suffixed versions are filtered out
+      expect(result).toBe('v25.3.1');
     });
 
-    it('should sort prereleases correctly against each other', () => {
+    it('should return null if only prerelease versions exist', () => {
       const files = [
         { src: { stem: 'redpanda-properties-v25.4.0-rc1' } },
         { src: { stem: 'redpanda-properties-v25.4.0-rc2' } },
@@ -201,8 +207,8 @@ describe('Properties JSON file detection', () => {
       ];
 
       const result = extractLatestPropertiesVersion(files);
-      // rc2 > rc1 > beta1
-      expect(result).toBe('v25.4.0-rc2');
+      // All are prereleases, so no valid versions
+      expect(result).toBeNull();
     });
   });
 });
@@ -305,7 +311,7 @@ describe('Edge cases', () => {
     expect(extractLatestConnectVersion(filterConnectFiles(connectFiles))).toBe('4.93.0');
   });
 
-  it('should handle files with similar names', () => {
+  it('should handle files with similar names and reject suffixed versions', () => {
     const files = [
       { src: { stem: 'redpanda-properties-v25.3.1', component: 'streaming', module: 'reference' } },
       { src: { stem: 'redpanda-properties-schema', component: 'streaming', module: 'reference' } },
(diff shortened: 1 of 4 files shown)


Feediver1 (commented):
## Docs standards review (Antora extension PR)

**Files reviewed:** 5 (1 extension change, 1 unrelated extension fix, 1 new test file, package.json/lock). This is an Antora extension PR (JS code), not an `.adoc` content PR — the usual docs standards don't apply. Review focuses on correctness, test quality, and cross-repo coordination.

### What this PR does

Teaches the `set-latest-version` extension to scan the content catalog for actual JSON attachments and set two new attributes that point to versions which physically exist:

- `available-properties-tag` — highest stable version of `redpanda-properties-*.json` found in `streaming:reference` (or legacy `ROOT:reference`).
- `available-connect-version` — highest stable version of `connect-*.json` found in `connect:components` (or legacy `redpanda-connect:components`).

These solve the build-log noise from `latest-redpanda-tag` / `latest-connect-version` pointing to GitHub releases that don't yet have published JSON. The companion UI PR (docs-ui#384) consumes these attributes via an `(or available-* latest-*)` fallback pattern.

Also included: a small unrelated fix in `replace-attributes-in-attachments.js` (Self-Managed → Streaming title comparison) and a 25-test unit file for the new filtering/sorting logic.

### Cross-repo coordination

Your docs-ui#384 description explicitly states both PRs should merge together. Respecting that instruction is the right call — landing one without the other would leave deployed code that looks like the fix shipped while builds continue to throw the same unresolved-resource errors, making triage harder.

### Critical issues

1. **DIRTY merge state.** `mergeable: CONFLICTING` — needs a rebase against `main` before merge. Looks straightforward given the small file footprint, but it's a blocker.

2. **Undocumented Self-Managed → Streaming change in `extensions/replace-attributes-in-attachments.js` (line ~181).** This one-line fix (`componentVersion.title === 'Self-Managed'` → `'Streaming'`) is a real change that piggybacks on this PR but isn't mentioned in the description. If the change later proves wrong, no one would think to look here when debugging dynamic-attribute replacements. Either split it into its own PR or add it to the "Changes" section of the body.

3. **CodeRabbit's test-duplication finding is still open.** `__tests__/extensions/version-fetcher/available-json-versions.test.js` re-implements `filterPropertiesFiles`, `filterConnectFiles`, `extractLatestPropertiesVersion`, and `extractLatestConnectVersion` as local copies of the extension's internal logic (lines 12–53). The tests will keep passing even if `set-latest-version.js` diverges from these copies — meaning the tests provide algorithmic coverage but not regression coverage of the actual code path. Fix: extract the four helpers to named functions in `set-latest-version.js`, export them, and `require()` them in the test. The behavioral coverage is already good; only the wiring is off.

### Suggestions

1. **Asymmetric `v` prefix handling between properties and connect.** `extractLatestPropertiesVersion` normalizes a leading `v` (`v.replace(/^v/, '')`), but `extractLatestConnectVersion` does not. If a future Connect JSON file is ever named `connect-v4.79.0.json`, it'd silently fail to be detected. Either normalize both for symmetry, or add a one-line comment to the connect filter noting that Connect tags never carry a `v` prefix (which is the current convention, per the `FALLBACK_VERSIONS = ['4.79.0', '4.78.0', ...]` array in the docs-ui gulp task).

2. **Package version bump 5.0.1 → 5.0.2 is a patch.** This PR adds two new public-facing attributes that other repos (docs-ui#384) depend on, plus a behavior change in `replace-attributes-in-attachments.js`. Per strict semver, that's a minor bump (5.1.0). Worth confirming with team convention — internal-tooling extensions sometimes treat versioning loosely.

3. **Document the legacy component-name fallbacks inline.** The filters accept `streaming` || `ROOT` and `connect` || `redpanda-connect`. The current vs legacy names aren't called out in the code — a one-line comment noting which is current ("`streaming` is the current name; `ROOT` is preserved for older playbooks") would help future maintainers know which fallback to remove first when the legacy name is retired.

4. **`logger.debug` for "No JSON files found in content catalog"** is appropriate, but if a misconfigured Antora playbook produces zero matches, this should arguably be `logger.warn` since it indicates a problem rather than normal absence. Up to team preference — flagging because the silent-fallback model relies on someone noticing the build-log absence.

5. **Test file lacks an end-to-end integration assertion.** The 25 tests cover the algorithm thoroughly, but none exercise the extension's `register` function with a fake `components` array to confirm that `asciidoc.attributes['available-properties-tag']` is actually set as a side effect. Combined with Critical 3, the test suite gives high confidence in the algorithm and low confidence in the integration.

### CodeRabbit findings

1. **Prerelease filter (line 154–156 of `set-latest-version.js`).** ✓ **Already addressed.** The follow-up commit added `semver.prerelease(normalized) === null` (properties side) and `semver.prerelease(v) === null` (connect side), correctly excluding suffixes like `-backup` / `-rc1`. Both filters now produce only stable releases.
2. **Test duplication.** ❌ Still open — flagged as Critical 3 above.

### What works well

- **Backwards-compatible component-name handling** preserves builds against older playbooks (`ROOT`, `redpanda-connect`).
- **Prerelease versions correctly excluded** after CodeRabbit's nudge, on both filter paths.
- **Clear separation of debug and info logging** — the `logger.info` summary at the end is genuinely useful operator-facing output.
- **The `if (versions.length > 0)` guards** prevent `asciidoc.attributes` from being set to `null` when no matching files exist.
- **The companion `replace-attributes-in-attachments.js` fix is a real catch** — just needs documentation.
- **The two-attribute design** (separate `available-properties-tag` and `available-connect-version`) keeps Connect's release cadence decoupled from Redpanda's. Good API shape.

Feediver1 (commented):
## Follow-up review on the delta

`@JakeSCahill` — re-reviewed after `f6cf9e0` and the rebase. **All three critical issues from my first review are resolved.**

| # | First-review critical | Status |
|---|---|---|
| 1 | DIRTY merge state | ✅ Resolved. `mergeable: MERGEABLE` now. Branch rebased cleanly. |
| 2 | Undocumented Self-Managed → Streaming change in `extensions/replace-attributes-in-attachments.js` | ✅ Resolved. That file is no longer in the diff — either split into a separate PR or absorbed via the rebase. The PR is now cleanly scoped to the `available-*` attributes work + tests. |
| 3 | Test duplication (tests re-implementing source logic) | ✅ Resolved. Commit `f6cf9e0` exports `filterPropertiesFiles`, `filterConnectFiles`, `extractLatestPropertiesVersion`, `extractLatestConnectVersion` from `set-latest-version.js`. The test now imports the real implementations via `require()` instead of duplicating them. Dependency-injecting `semver` via the wrapper helpers is a clean way to keep the test deterministic. Real-code-path coverage now in place. |

CI is fully green across the matrix (test-extensions, test-macros, test-tools, test-asciidoc-extensions, test-property-extractor 3.9 + 3.11, test-all, MCP integration, CGO E2E). Ready for human approval.

### Open from first review (none blocking)

- **`v` prefix asymmetry** between properties and connect filters — fine if Connect tags never carry `v`, optional one-line comment.
- **`logger.debug` for "No JSON files found"** — team preference; debug is acceptable.
- **No end-to-end `register()` integration test** — Critical 3's fix delivers most of the value; full register-side-effect coverage would be a nice-to-have.

### Cross-PR coordination

This PR bumps `5.0.2 → 5.0.3`, and #201 also bumps `5.0.2 → 5.0.3` (same target). Whichever PR lands second will hit a merge conflict on `package.json`. Worth coordinating the merge order — happy either way as long as one of them rebases after the other lands.

Feediver1 (approved):
lgtm

PR #201, opened 2026-06-01
fix: connector docs automation fixes (false positives + prometheus {})

JakeSCahill:
## Summary

This PR fixes two issues in the connector docs automation:

### 1. False Positive "New Connector" Reports

**Problem**: GitHub Actions run [26747630215](https://github.com/redpanda-data/rp-connect-docs/actions/runs/26747630215) reported 5 "new" connectors that weren't actually new:
- `tigerbeetle_cdc` (inputs) - CGO-only
- `zmq4` (inputs/outputs) - CGO-only  
- `ffi` (processors) - CGO-only
- `a2a_message` (processors) - OSS connector

**Root Cause**: In intermediate version processing, `oldData` was loaded from potentially stale disk files without CGO augmentation, while `newData` was freshly augmented with binary analysis. This asymmetric comparison caused CGO connectors to appear "new".

**Fix**: 
- Force re-fetch both old and new version data (`forceFresh: true`)
- Run binary analysis on BOTH versions
- Augment BOTH before comparison for symmetric diff

### 2. Prometheus `{}` Missing in Common Config YAML

**Problem**: PR [#415](https://github.com/redpanda-data/rp-connect-docs/pull/415) fixed missing `{}` brackets for prometheus, but the fix was overwritten by subsequent automation runs.

**Root Cause**: `buildConfigYaml.js` didn't add `{}` when all component fields are advanced (no fields rendered in "common" mode), producing invalid YAML:
```yaml
metrics:
  prometheus:
```

**Fix**: Detect when no fields will be rendered and add `{}` for valid YAML:
```yaml
metrics:
  prometheus: {}
```

## Testing

- Added 4 tests for symmetric augmentation fix
- Added 2 tests for prometheus `{}` scenario
- All existing tests pass

## Related

- rp-connect-docs PR #435 (updated with correct content)
- Upstream FFI glibc issue: https://github.com/redpanda-data/connect/issues/4471

### DIFF
--- a/__tests__/tools/buildConfigYaml.test.js
+++ b/__tests__/tools/buildConfigYaml.test.js
@@ -254,5 +254,43 @@ describe('buildConfigYaml', () => {
       // Array-of-objects with empty children should render as empty array
       expect(result).toContain('items: []');
     });
+
+    it('should render {} when all fields are advanced and includeAdvanced is false (prometheus scenario)', () => {
+      // This mimics the prometheus metrics component where all fields are advanced
+      const allAdvancedFields = [
+        {
+          name: 'use_histogram_timing',
+          type: 'bool',
+          kind: 'scalar',
+          description: 'Use histogram for timing.',
+          is_advanced: true,
+          default: false
+        },
+        {
+          name: 'push_url',
+          type: 'string',
+          kind: 'scalar',
+          description: 'Push gateway URL.',
+          is_advanced: true,
+          default: ''
+        }
+      ];
+
+      const result = buildConfigYaml('metrics', 'prometheus', allAdvancedFields, false);
+      // When no fields to render in common mode, should use {} for valid YAML
+      expect(result).toContain('metrics:');
+      expect(result).toContain('  prometheus: {}');
+      // Should NOT have a trailing colon without value
+      expect(result).not.toMatch(/prometheus:\s*$/m);
+    });
+
+    it('should NOT render {} when fields are present', () => {
+      const result = buildConfigYaml('metrics', 'prometheus', sampleChildren, false);
+      // When fields are present, should have a colon without {}
+      expect(result).toContain('metrics:');
+      expect(result).toContain('  prometheus:');
+      expect(result).not.toContain('prometheus: {}');
+      expect(result).toContain('    host:');
+    });
   });
 });
--- a/__tests__/tools/connector-diff-cgo.test.js
+++ b/__tests__/tools/connector-diff-cgo.test.js
@@ -380,3 +380,147 @@ describe('CGO Connector Diff - Deprecation Detection', () => {
     expect(diff.summary.deprecatedFields).toBe(1);
   });
 });
+
+describe('CGO Connector Diff - Symmetric Augmentation (False Positive Fix)', () => {
+  // These tests verify the fix for false "new connector" reports
+  // caused by asymmetric augmentation in intermediate processing
+
+  test('should NOT report CGO connectors as new when both old and new are augmented', () => {
+    // Scenario: Both versions are properly augmented with CGO connectors
+    // This is the expected state after the fix
+    const cgoConnectors = [
+      { name: 'tigerbeetle_cdc', type: 'input', status: 'stable', requiresCgo: true, config: { children: [] } },
+      { name: 'zmq4', type: 'input', status: 'stable', requiresCgo: true, config: { children: [] } },
+      { name: 'ffi', type: 'processor', status: 'stable', requiresCgo: true, config: { children: [] } }
+    ];
+
+    const oldIndex = {
+      inputs: cgoConnectors.filter(c => c.type === 'input'),
+      processors: cgoConnectors.filter(c => c.type === 'processor')
+    };
+
+    const newIndex = {
+      inputs: cgoConnectors.filter(c => c.type === 'input'),
+      processors: cgoConnectors.filter(c => c.type === 'processor')
+    };
+
+    const diff = generateConnectorDiffJson(oldIndex, newIndex, {
+      oldVersion: '4.93.0',
+      newVersion: '4.94.0'
+    });
+
+    // CGO connectors should NOT be reported as new
+    expect(diff.summary.newComponents).toBe(0);
+    expect(diff.details.newComponents).toHaveLength(0);
+  });
+
+  test('should detect genuinely new connectors when both versions are augmented', () => {
+    // Scenario: Both versions augmented, but new version has an additional new connector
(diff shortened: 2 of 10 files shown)


Feediver1 (commented):
## Docs standards review (Antora extension / tooling PR)

**Files reviewed:** 10 files (3 new test files, 2 JS source files, 3 Python source files, package.json/lock). Not an `.adoc` content PR — review focuses on correctness, scope, and cross-PR coordination.

### What this PR does

The PR title and body describe two JS fixes:

1. **Symmetric augmentation for connector diff** (`rpcn-connector-docs-handler.js`, `connector-diff-cgo.test.js`). Force-fresh data loading for both old and new versions, run binary analysis on both, augment both before comparison. Eliminates false-positive "new connector" reports for CGO-only connectors (`tigerbeetle_cdc`, `zmq4`, `ffi`, `a2a_message`) that were appearing in intermediate diff runs.
2. **Prometheus `{}` in common config YAML** (`buildConfigYaml.js`, `buildConfigYaml.test.js`). Pre-filter children into a `fieldsToRender` array; if it's empty, emit `connector: {}` instead of a dangling `connector:` which produced invalid YAML.

But the diff also contains a **third, unmentioned change**:

3. **Python numeric-constant resolution** (`constant_resolver.py`, `property_extractor.py`, `transformers.py`, `constant-resolver.test.js`). Adds `_parse_size_literal()` (256_MiB, 20_GiB, etc.) and `resolve_numeric_constant()` to the property extractor, and threads `constant_resolver` through the `FriendlyDefaultTransformer` constructor. The new test file's own header explicitly says: *"Related to DOC-2137: Property docs should show resolved values instead of constant names."*

### Critical issues

None. Every check in the CI matrix passes (test-extensions × 2 Node versions, test-tools × 2, test-macros × 2, test-property-extractor × 2 Python versions, CGO detection E2E, MCP integration, Netlify, CodeRabbit).

### Scope concern

**Item 3 (the Python constant-resolver work) is bundled into a PR titled "connector docs automation fixes."** This is the same pattern as docs-ui#385 today — multiple unrelated fixes in one PR. The Python work belongs to DOC-2137, a separate ticket. Bundling has the same downsides as before: if anything regresses, reverting takes back the unrelated work; future archaeologists looking for "DOC-2137" by PR title won't find this PR; the test plan in the PR body only covers items 1 and 2.

This is a soft call — the changes are all in the connector-docs tooling and all small — but at minimum the PR description should be updated to enumerate the three changes, and DOC-2137 should be referenced so the ticket and the work are linked.

### Suggestions

1. **CodeRabbit's `require()`-in-loop nit (line 879–880 of `rpcn-connector-docs-handler.js`).** Two `require()` calls for `analyzeAllBinaries` and `findCloudVersionForDate` are currently inside the `for` loop that iterates over intermediate releases. Module caching makes the performance impact negligible, but hoisting them above the loop is cleaner and matches the rest of the file's import-at-top convention. Trivial fix.

2. **Source-tree pollution risk in `constant-resolver.test.js`.** `beforeAll()` writes `test_constant_resolver.py` into `tools/property-extractor/` and `afterAll()` cleans it up — good defensive coding. But if Jest is killed mid-suite (Ctrl+C, OOM, runner timeout), the stray file persists in the source tree and the next test run may behave unexpectedly. `os.tmpdir()` is already imported at the top — moving the script there would eliminate the risk entirely.

3. **Cross-language test design.** The Python `ConstantResolver` is tested via a JS Jest test that spawns Python through `execSync` and parses JSON output. It works (CI is green across `test-property-extractor (3.9)` and `(3.11)`), but couples the test to Node + Python version assumptions and shells out for each assertion. The `test-property-extractor` CI matrix already runs pytest under both Python versions — moving this test to a `pytest` file under `tools/property-extractor/tests/` would let Python developers run it directly with native introspection, no subprocess fragility, and no JSON marshaling round-trip.

4. **`fieldsToRender` refactor is genuinely cleaner** than the original conditional inside the `forEach`. ✓

5. **Console log bullets changed from `•` to `-`.** Cosmetic, but if any CI summary parser greps for `   •` to extract counts, this is a silent behavior change. Probably no consumers, but worth a quick check.

### Cross-PR coordination concern

**Both this PR and #200 bump `package.json` version.** Reviewed status:

- PR #200 (the `available-*` attributes PR): bumps 5.0.1 → 5.0.2. **Currently DIRTY/CONFLICTING against main.**
- PR #201 (this one): bumps 5.0.2 → 5.0.3.

This PR's bump *assumes* #200 has already landed. If this PR merges first, main jumps to 5.0.3 and #200's "5.0.1 → 5.0.2" bump becomes nonsensical (and will conflict). If #200 merges first, this PR's 5.0.2 → 5.0.3 is correct.

Either coordinate the merge order (recommended: #200 → #201) or rebase whichever lands second to skip the version step.

### Behavioral change to watch

**`FriendlyDefaultTransformer.__init__` now requires a `constant_resolver` argument** (per the `transformers.py` and `property_extractor.py` diff). If any other callsite constructs `FriendlyDefaultTransformer` without the new arg, that callsite will break at runtime. The diff updates `property_extractor.py:723` (the only caller I can see in the diff) but worth a quick `grep -r FriendlyDefaultTransformer tools/` to confirm nothing else constructs it.

### CodeRabbit findings worth considering

1. **`require()` in loop** — see Suggestion 1.

### What works well

- **Root-cause fix, not symptom mask.** Symmetric augmentation directly addresses *why* CGO connectors were appearing as new (asymmetric data sourcing). The fix is local to the diff-generation step and the test (`'should NOT report CGO connectors as new when both old and new are augmented'`) directly proves the behavior.
- **`forceFresh: true` opt-in parameter** keeps the existing callers unaffected. Other code paths that rely on the disk cache continue to work.
- **`fieldsToRender` pre-filter in `buildConfigYaml.js`** is both cleaner and slightly more efficient (one filter pass vs N conditional skips inside the forEach).
- **Symmetric release-info lookup** (now resolving cloud version for both `fromVer` and `toVer`) is exactly what the false-positive fix needs — both sides of the diff get the same treatment.
- **Test coverage targets the exact regression** — there's a named test for the false-positive scenario and a named test for the prometheus `{}` scenario. Tests will fail loudly if either fix regresses.
- **The Python `_parse_size_literal` handles both binary and decimal suffixes** (`_MiB` vs `_MB`, `_GiB` vs `_GB`) which is important for C++ constants that may use either convention. The implementation also gracefully falls back to plain int parsing.
- **`afterAll` cleanup is present** on the dynamic test script — see Suggestion 2 for the further hardening.

Feediver1 (approved):
No critical issues. Suggestions are your call.

PR #204, opened 2026-06-12
fix(property-docs): improve constant resolution logging and reliability

JakeSCahill:
## Summary

Fixes constant resolution in property documentation automation. Previously, properties like `topic_memory_per_partition` showed unresolved constant names (e.g., `DEFAULT_TOPIC_MEMORY_PER_PARTITION`) instead of actual values (e.g., `200 KiB (204800)`).

## Changes

- Enhanced logging for ConstantResolver initialization (DEBUG → INFO level)
- Added warning messages when Redpanda source is not found or inaccessible
- Improved constant resolution error messages for better debugging
- Added detailed logging for successful and failed constant resolutions

## Result

The constant `DEFAULT_TOPIC_MEMORY_PER_PARTITION` now correctly resolves to:
```json
{
  "default": "200 KiB (204800)"
}
```

Instead of:
```json
{
  "default": "DEFAULT_TOPIC_MEMORY_PER_PARTITION"
}
```

## Testing

Verified by running property extraction:
```bash
python3 property_extractor.py --recursive --path tmp/redpanda --output /tmp/test-properties.json
```

Confirmed that `topic_memory_per_partition` property now shows the resolved value.

## Version Bump

Bumped package version from 5.1.0 to 5.1.1.

## JIRA

https://redpandadata.atlassian.net/browse/DOC-2137

---


JakeSCahill:
@coderabbitai review

kbatuigas (commented):
@JakeSCahill Please see:

> Findings
> 
> 1. High: the default property-docs command no longer generates AsciiDoc output unless `GENERATE_PARTIALS=1` is set.
> The new gate in Makefile only runs `generate-docs` when `GENERATE_PARTIALS=1`, and Makefile makes the target itself a no-op otherwise. But the main CLI wrapper only sets that env var when the user explicitly passes `--generate-partials` in doc-tools.js. That means `npx doc-tools generate property-docs --tag v25.3.1` now skips doc generation entirely, despite the documented contract in README.adoc and README.adoc, which still present the default command as generating docs.
> Impact:
> - existing non-`--generate-partials` workflows silently degrade to JSON-only output
> - callers that rely on the default command behavior will stop updating generated docs
> 
> 2. Medium: the “improved reliability” change still ignores the extractor’s explicit source path and can disable constant resolution incorrectly.
> The new initialization/warning path in property_extractor.py still discovers Redpanda sources via `find_redpanda_source()`, which only searches a few hardcoded temp locations in property_extractor.py. The actual CLI source path is the user-supplied `--path`, passed later as `options.path` in property_extractor.py. So if someone invokes the extractor against any nonstandard checkout path, the new code will now warn that constant resolution is disabled and leave defaults unresolved even though the source path was provided correctly.
> Impact:
> - custom-path invocations still fail to resolve constants
> - the new warnings can be misleading because they report a missing source tree when the source was actually supplied via CLI
> 
> Open questions
> 
> 1. Was the Makefile change meant to optimize away only partial generation, or to suppress all AsciiDoc generation unless `--generate-partials` is passed? The current CLI/docs contract points to the former.
> 2. Should constant resolution be initialized from `options.path` instead of a second filesystem probe? That would match the actual extractor input and make the new warnings accurate.

PR #205, opened 2026-06-17
DOC-1878: Chunk long reference pages for Algolia indexing

JakeSCahill:
## What

Splits long reference pages (cluster/topic properties, metrics) into multiple Algolia records in `extensions/algolia-indexer/generate-index.js`, so each record's `titles` array stays small, and adds a clean `url` attribute (base page path) to every record.

- `chunkTitles()` groups headings into ~180-word chunks; the first chunk keeps the page's base `objectID`, later chunks use `objectID = base#<firstHeadingAnchor>`.
- `url` = base page path (no `#fragment`) — used by the search UI for result links and as the attribute to de-duplicate chunks on (`attributeForDistinct`).
- New tests in `__tests__/extensions/algolia-indexer.test.js`.

## Why

Investigation for [DOC-1878](https://redpandadata.atlassian.net/browse/DOC-1878). The **primary** root cause is an Algolia index setting (`separatorsToIndex` includes `_`, which breaks exact underscored-name search on large pages) — fixed separately in the Algolia dashboard, not in this repo.

This change is **defense-in-depth / record-size hygiene**: it keeps records small enough that the per-record indexing cliff can't recur (e.g. if `topic-properties` ever grew large). Safe to land independently; not required to fix the bug on its own.

## Testing

- `npm test -- __tests__/extensions/algolia-indexer.test.js` — 4 new tests pass.
- Validated against real index data: the 429-entry cluster-properties page chunks into 11 records with every name preserved.

[DOC-1878]: https://redpandadata.atlassian.net/browse/DOC-1878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/__tests__/extensions/algolia-indexer.test.js
+++ b/__tests__/extensions/algolia-indexer.test.js
@@ -1,0 +1,132 @@
+'use strict'
+
+const generateIndex = require('../../extensions/algolia-indexer/generate-index.js')
+
+const noopLogger = { info: () => {}, warn: () => {}, error: () => {}, debug: () => {} }
+
+// Build HTML mirroring Antora's render of a doc page: an `article.doc` with an
+// h1, an intro paragraph, and a set of section headings (with ids) plus body text.
+function buildArticle ({ h1 = 'Reference', intro = 'Intro paragraph.', sections = [] }) {
+  const body = sections
+    .map(({ name, level = 'h3' }) =>
+      `<div class="sect2"><${level} id="${name}">${name}</${level}>` +
+      `<div class="paragraph"><p>Description for ${name}.</p></div></div>`)
+    .join('')
+  return `<article class="doc"><h1>${h1}</h1>` +
+    `<div class="paragraph"><p>${intro}</p></div>${body}</article>`
+}
+
+const component = { name: 'redpanda', title: 'Self-Managed', latest: { version: '25.3' } }
+
+function makePage (html, overrides = {}) {
+  return {
+    contents: Buffer.from(html),
+    out: { dirname: 'current/reference', basename: 'page.html' },
+    pub: { url: '/current/reference/page' },
+    src: { component: 'redpanda', version: '25.3', origin: {} },
+    asciidoc: { attributes: {} },
+    ...overrides
+  }
+}
+
+function runIndex (page) {
+  const result = generateIndex(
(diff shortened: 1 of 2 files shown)


micheleRP (approved):
Approving — the chunking logic is clean, the tests mirror the tokenizer well, and I confirmed stale chunk records are reconciled/deleted on rebuild (`extensions/algolia-indexer/index.js:131-152`), so churn from changing headings is handled.

Two notes for your consideration:

1. **Deploy coordination (most important).** Chunks share the same `title`, intro-derived `text`, and `url`, so a query that matches the page title/intro will match *every* chunk of a page. The comments reference `attributeForDistinct: url` for de-duplication, but that's a dashboard-only setting (not in this repo). Since the index regenerates on every production build, could you confirm `distinct: true` + `attributeForDistinct: url` is set in the Algolia dashboard before these chunk records reach production? Otherwise search may show duplicate page entries (this also affects any non-properties page with >30 headings, which `isPropertiesPage` chunks too).

2. **Variable shadowing (nit).** In the `titleChunks.forEach((chunkTitles, ...) =>` callback, the param `chunkTitles` shadows the `chunkTitles()` function declared below. Works today; renaming the param (e.g. `chunk`) would avoid a future foot-gun.

JakeSCahill:
Thanks Michele! Addressed both notes:

**2. Variable shadowing (nit)** — Fixed in 1e39bf1c: renamed the `titleChunks.forEach` callback param from `chunkTitles` to `chunk` so it no longer shadows the `chunkTitles()` function. Tests still pass.

**1. Deploy coordination** — Confirmed: `distinct: true` + `attributeForDistinct: url` is a dashboard-only setting. The indexer extension (`extensions/algolia-indexer/index.js`) only pushes records via `multipleBatch` and never calls `setSettings`, so there's nothing in this repo that configures it. We'll keep it dashboard-managed and ensure `distinct: true` + `attributeForDistinct: url` is set in the Algolia dashboard **before** these chunk records reach production, to avoid duplicate page entries (including any non-properties page with >30 headings, which `isPropertiesPage` also chunks). Flagging as a pre-merge/deploy coordination item rather than a code change.

PR #207, opened 2026-07-06
fix(rpk-docs): pin Docker Go image to go.mod version and fix schema validator path

JakeSCahill:
## Summary

Two fixes for the `generate rpk-docs` command discovered while running the automation against rpk v26.1.12.

### Docker Go version mismatch

`golang:1` resolves to whatever the latest 1.x patch release is on Docker Hub, which can lag behind the version required by rpk's `go.mod`. When there's a mismatch, Go refuses to build with `GOTOOLCHAIN=local`:

```
go: go.mod requires go >= 1.26.4 (running go 1.26.3; GOTOOLCHAIN=local)
```

The error was caught by the fallback handler but the message was only printed to stderr — making it look like a silent failure and causing the generator to fall back to native build (no plugin installation, missing plugin subcommands in the tree).

**Fix:** Read the required version from `go.mod` via the existing `getRequiredGoVersion()` helper and use `golang:<version>` so the container image always satisfies the source requirement.

### Schema validator path

`loadSchemaValidator()` resolved the schema path relative to `__dirname` (inside `node_modules`), but the schema file (`rpk-overrides.schema.json`) lives next to the overrides file in the consuming docs repo. The validator always threw `Schema file not found` and silently skipped schema validation.

**Fix:** Look for the schema as a sibling to the overrides file first (the common case), falling back to the package-relative path for monorepo/legacy usage. `overridesPath` is threaded from `loadAndValidateOverrides` → `validateOverrides` → `validateSchema` → `loadSchemaValidator`.

## Test plan

- [ ] Run `npx doc-tools generate rpk-docs --ref v26.1.12` with Docker running — Docker build should succeed with `golang:1.26.4` and all plugins should install
- [ ] Run with an overrides file that has a sibling schema — schema validation should pass instead of throwing "Schema file not found"
- [ ] Run without a schema file present — should fall back gracefully to the package-relative path (or throw the existing error if neither exists)


kbatuigas (commented):
Mind checking if these are valid?

> - tools/rpk-docs/generate-rpk-docs.js:2538 uses shouldUsePartialDir(overrides, commandPath) when deciding whether to write a file into cloudSecretDir, but the rest of the generator uses resolvedOverrides for the same decision (2354-2358, 2608). That creates a real mismatch for any asPartial that arrives through $ref/$refs: nav/subcommand xrefs will treat the command as a partial and omit links, while generation still writes the page into the normal output tree. This will leave orphaned/generated pages in the wrong location. The write-path check should use resolvedOverrides too.
> 
> - tools/rpk-docs/generate-rpk-docs.js:2579-2588 does not fully implement the claimed stale-file cleanup. It only scans directories that received at least one new write, and it explicitly skips the root output dirs. Two cases still leak stale docs: removed/excluded root-level command pages like rpk-help.adoc, and removed/excluded subtrees where an entire managed subdirectory now has zero surviving files. In both cases, no deletion pass ever reaches those files, so the repo keeps outdated pages.
> 
> The targeted Jest suites passed, but they do not cover either of the cases above. I ran:
> npm test -- __tests__/tools/rpk-docs/generate-rpk-docs.test.js __tests__/tools/rpk-docs/override-features.test.js __tests__/tools/rpk-docs/rpk-docs-handler.test.js
> The test run then failed only at the end because jest-junit tried to write test-results/junit.xml outside the writable sandbox.

JakeSCahill:
Thanks @kbatuigas — I checked both and they're valid. Fixed in ef642763:

**1. Partial-routing mismatch (write path vs. nav/xref).** The write-path decision at `generate-rpk-docs.js` used `shouldUsePartialDir(overrides, …)` (raw), while nav (2138) and the subcommand xref filter (2357) use `resolvedOverrides`. Since `resolveReferences` materializes `$ref`/`$refs`, an `asPartial` arriving via a `$ref` was invisible to the raw object — nav/xrefs treated the command as a partial and dropped its links, while generation still wrote the page into the normal output tree. Changed the write path to `resolvedOverrides` so both agree.

**2. Incomplete stale-file cleanup.** Confirmed both leaks: (a) removed/excluded root-level pages like `rpk-help.adoc` were never scanned (root dirs were skipped), and (b) fully-removed subtrees contributed zero writes, so their subdir never entered the scan set. Rewrote cleanup to scan `outputDir` + all `rpk-*` subdirs of both `outputDir` and `cloudSecretDir`, deleting only files matching `rpk(-*)?.adoc` so hand-written partials at the shared `partials/` root are never touched, and removing now-empty `rpk-*` subdirs.

Added integration tests for both findings (each verified to fail without its fix). Full rpk-docs suite: 286/286 passing.

tools/rpk-docs/generate-rpk-docs.js
```diff
@@ -1997,6 +2082,90 @@ function findTopLevelWithSubcommands(tree) {
   return result
 }
 
+/**
+ * Static entries that always appear at the top of the rpk nav section,
+ * before the auto-generated command entries. These are hand-written pages
+ * that are not part of the CLI command tree.
+ */
+const STATIC_RPK_NAV_ENTRIES = [
+  '*** xref:reference:rpk/rpk.adoc[]',
```
kbatuigas: Do we actually have an rpk/rpk.adoc?

kbatuigas (approved):
Just one more thing to look at - but I'll approve so I'm not blocking the PR

tools/rpk-docs/generate-rpk-docs.js
JakeSCahill: Good catch — no, we don't. I checked `redpanda-data/docs` on `main`: `modules/reference/pages/rpk/` has no `rpk.adoc`; the section landing page is the hand-written `index.adoc` (`:page-layout: index`). `rpk-commands.adoc` and `rpk-x-options.adoc` do exist and are hand-written.

Fixed in 80c75722: the generator now **skips the root `rpk` command** (mirroring `updateNavFile`, which already skipped it) rather than emitting `rpk.adoc`, and I dropped the `rpk.adoc` static nav entry. The root is represented by the existing hand-written `index.adoc`, which generation never touches.

While confirming this I also caught a related bug in the earlier stale-file cleanup on this PR: it scanned the root output dir and deleted any `rpk-*.adoc` not regenerated that run, which would have deleted the hand-written `rpk-commands.adoc`/`rpk-x-options.adoc` on every run. Fixed in 2d8b3501 by restricting cleanup to the generator-owned `rpk-*` subdirectories and never touching the root. Added tests for both (root command skipped + index.adoc untouched; root dir never cleaned). Full rpk-docs suite: 287/287.

PR #208, opened 2026-07-07
fix(rpcn-docs): whats-new xref truncation and config-component cloud metadata

JakeSCahill:
## Summary

Fixes three generator bugs observed in the auto-docs output for redpanda-data/rp-connect-docs#452 (the 4.96.3 → 4.98.0 update):

### 1. whats-new descriptions truncated by xref targets

`capToTwoSentences` in `rpcn-connector-docs-handler.js` split sentences on the periods inside Antora xref targets (`try.adoc`, `catch.adoc`), and `String.match` silently discarded all leading text that never formed a complete sentence match. The `try_catch` whats-new entry rendered as the fragment:

> adoc[`catch`] processors into a single block with an explicit recovery path.

**Fix:** protect `xref:...[...]` macros and bare doc/config filenames with placeholders before sentence splitting (same mechanism as the existing URL protection), and merge any dropped leading fragment into the first sentence as a general hardening.

### 2. config components missing cloudSupported (self-contradictory diff JSON)

`augmentConnectorData` stamps `cloudSupported`/`requiresCgo` using a hardcoded `connectorTypes` list that omitted `'config'`, so the new `error_handling` config component was emitted with `cloudSupported: false` in its component entry while `binaryAnalysis.details.cloudSupported` (which is type-agnostic) listed it as cloud-supported — flagged by CodeRabbit on the rp-connect-docs PR.

**Fix:** add `'config'` to the augmentation list and the no-binary-analysis fallback list, and strip platform metadata from `config` entries in the clean-OSS copy (they have no `config`/`fields` wrappers, so they bypass the existing filter).

### 3. Status column showed "object" for status-less components

`report-delta.js` fell back to `raw.type` when a component had no `status`, so the whats-new Status column and PR summary showed the type `object` for `error_handling`. Removed the type fallback and made PR summaries/gap reports render `(type)` instead of `(type, )` when status is empty.

## Testing

- New suite `__tests__/tools/rpcn-connector-summaries.test.js` (8 tests) covering all three fixes, including the exact `try_catch` description regression.
- Full run: 30 suites, 674 tests, all passing.

### DIFF
--- a/__tests__/tools/rpcn-connector-summaries.test.js
+++ b/__tests__/tools/rpcn-connector-summaries.test.js
@@ -18,7 +18,7 @@
 // Jest); stub it out since these tests never touch GitHub.
 jest.mock('../../cli-utils/octokit-client', () => ({}));
 
-const { capToTwoSentences, augmentConnectorData } = require('../../tools/redpanda-connect/rpcn-connector-docs-handler');
+const { capToTwoSentences, augmentConnectorData, buildCleanOssData } = require('../../tools/redpanda-connect/rpcn-connector-docs-handler');
 const { generateConnectorDiffJson } = require('../../tools/redpanda-connect/report-delta');
 
 describe('capToTwoSentences - xref and filename protection', () => {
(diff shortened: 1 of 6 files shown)


kbatuigas (commented):
One finding:

> [rpcn-connector-docs-handler.js (line 1127)](/private/tmp/tools/redpanda-connect/rpcn-connector-docs-handler.js:1127): the new config support is incomplete in the “clean OSS snapshot” path. augmentConnectorData() now adds augmentation-only config entries from cloudOnly/cgoOnly, but cleanData.config is only passed through stripPlatformMetadata() and never filtered back to OSS-only entries. On a subsequent run, any previously augmented cloud-only or cgo-only config component will be copied into ._connect-<version>-clean.json, so binary analysis is no longer comparing pure OSS data. That can mask self-hosted/cloud deltas or make them sticky across runs. The non-config types already avoid this by filtering to entries with config/fields; config needs an equivalent way to drop augmentation-only entries before analysis.

JakeSCahill:
Thanks @kbatuigas — valid, and fixed in 80829bc9.

Confirmed the leak is real across runs: `augmentConnectorData()` adds cloud-only/cgo-only `config` entries and the handler persists the augmented data back to `connect-<version>.json` (`fs.writeFileSync(dataFile, augmentedData)`), so the next run reloads them. The clean-snapshot path filtered standard types to `c.config || c.fields` (dropping augmentation-only entries) but the `config` branch only stripped platform metadata — so previously-augmented cloud-only/cgo-only config components got copied into `._connect-<version>-clean.json`, and binary analysis stopped comparing pure OSS data.

Config entries have no `config`/`fields` wrapper (genuine OSS entries carry `kind`/`description`/`children`), so I couldn't reuse that filter. Instead I drop augmentation-only config entries by their platform markers — `cloudOnly === true || requiresCgo === true`. That's safe because augmentation only ever sets `cloudOnly` on *added* entries, and a component present in the standard OSS build can't be cgo-only, so genuine OSS config entries always survive (verified against a real `connect-*.json` fixture: all 17 config entries have neither marker).

Also extracted the snapshot logic into `buildCleanOssData()` and added unit tests covering config drop, metadata strip, standard-type behavior, and a raw (un-augmented) index — the config-drop test was verified to fail without the fix. Full rpcn suites: 35/35 passing.

kbatuigas (approved):

PR #209, opened 2026-07-10
fix(cli): make rpk binary download resilient to transient errors

JakeSCahill:
## Problem

The \`Update RPCN Connector Docs\` workflow in \`rp-connect-docs\` intermittently fails while downloading the \`rpk\` binary during its **Install documentation tools** step (which runs \`doc-tools install-test-dependencies\`).

The root cause is in \`cli-utils/install-test-dependencies.sh\` → \`install_rpk()\`, which fetched the release zip with a single, fragile call:

\`\`\`bash
if curl -LO "$rpk_url"; then
\`\`\`

Three compounding issues:

1. **No retry** — one attempt only. Any transient GitHub/CDN 5xx, connection reset, TLS/DNS blip, or \`latest\` redirect hiccup fails the whole step.
2. **No \`--fail\`** — \`curl\` exits \`0\` on HTTP 403/404/429/503 and writes the error page into the zip. The subsequent \`unzip ... 2>/dev/null\` then fails with the misleading \`"may not exist for linux-amd64"\` message, masking the real cause — often unauthenticated GitHub **rate limiting** on CI runners.
3. **No timeouts** — a stalled connection can hang until the job cap.

This was also inconsistent within the same repo: the downloads in \`cli-utils/dependencies.js\` already use the fail-fast pattern. (The \`buf\` download in this same script had the identical single-attempt fragility — now fixed here too.)

## Fix

\`\`\`bash
curl -fL --retry 5 --retry-all-errors \
     --connect-timeout 30 --max-time 300 --retry-max-time 600 \
     -o "$rpk_filename" "$rpk_url"
\`\`\`

- \`-f\` — HTTP >= 400 fails instead of producing a corrupt zip.
- \`--retry 5 --retry-all-errors\` — ride out transient 5xx/429/network errors, using curl's default exponential backoff (1s, 2s, 4s, ...) so retries can outlast GitHub secondary rate-limit windows.
- \`--connect-timeout\` / \`--max-time\` — bound hangs; \`--retry-max-time 600\` caps the whole retry window so persistent failures fail fast.
- Cleans up a possible partial download in the failure branch (\`-f\` stops error pages, not mid-transfer drops).
- Applies the same hardening to the \`buf\` download in this script.

_Review feedback from #209's code review is addressed in the second commit: exponential backoff instead of \`--retry-delay 3\`, failure-path cleanup, total retry-window cap, corrected description, and the \`buf\` follow-up._

## Verification

- \`bash -n\` passes.
- Real download succeeds and \`unzip -t\` reports "No errors detected".
- Forced HTTP 404 now retries then fails cleanly (exit 56, **no file written**) instead of falling through to the confusing unzip error.


Feediver1 (commented):
## Code review

**Verdict:** Correct, well-scoped fix with an accurate root-cause analysis. Nothing blocking — one valid cleanup nitpick (CodeRabbit's), one retry-tuning suggestion, and one inaccuracy in the PR description worth correcting.

### Correctness

- **`-o "$rpk_filename"` is equivalent to the old `-LO`** — the URL ends in `${rpk_filename}`, so the on-disk name is unchanged and the downstream `unzip`/cleanup paths still match. The explicit form is an improvement (no dependency on the URL's last segment).
- **`--retry-all-errors` is the right call here**, not an overreach: plain `--retry` does not retry 403s, and GitHub's unauthenticated rate limiting (the suspected main culprit) commonly presents as 403. Tradeoff: a genuine 404 (artifact truly missing for a platform) now takes ~5 retries to fail instead of failing fast — acceptable, and the existing warn message covers that case.
- The `command_exists rpk` guard at the top of the function is intact, so the retry loop only runs on cold installs.

### Suggestions

1. **`--retry-delay 3` disables curl's exponential backoff and may defeat the rate-limit goal.** Without `--retry-delay`, curl backs off 1s, 2s, 4s, 8s, 16s (~31s total window); with it, you get five fixed 3-second delays (~15s total). GitHub secondary rate limits typically need tens of seconds to clear, so all five retries can burn inside the same limit window. Dropping `--retry-delay 3` (keeping default exponential backoff) is both simpler and more likely to ride out the exact failure this PR targets.
2. **Add cleanup on the download-failure path** (CodeRabbit's nitpick — valid). `-f` prevents writing HTTP *error pages*, but a mid-transfer connection drop after a 200 can still leave a partial `$rpk_filename` in the working directory, and the `else` branch is the only failure path in this function that doesn't `rm -f "$rpk_filename"`. Trivial severity: a leftover partial zip gets overwritten by `-o` on re-run, so this is tidiness, not a bug.
3. **Optional: cap the total retry window.** `--max-time 300` bounds each *attempt*, so the pathological worst case (stalled-but-alive connections) is ~6 × 300s ≈ 30 minutes before failure. `--retry-max-time 600` would bound the whole operation if fast CI failure matters.

### PR description accuracy

The body claims "`buf`, `crd-ref-docs`, and `helm-docs` downloads already use fail-fast (`-f`)" — that doesn't hold for this script: the `buf` download (line 107) uses `curl -sSL` (no `-f`, no retry), and `helm-docs`/`crd-ref-docs` don't download anything here (they print manual-install instructions). The fail-fast pattern this PR matches actually lives in `cli-utils/dependencies.js`. Worth fixing the sentence — and it reveals a follow-up: **the `buf` download has the identical fragility this PR fixes for rpk** and would benefit from the same treatment.

### Security / tests

- No change in security posture: HTTPS GitHub releases URL, still no checksum verification — pre-existing and consistent with the rest of the script; `releases/latest` is inherently unpinned anyway.
- No tests, but the repo has no shell-test harness and this is CI-exercised by nature; fine.

_🤖 Automated code review via docs-team-standards._

Feediver1 (approved):

PR #210, opened 2026-07-16
feat(connect): extract connector metadata sections into regenerated partials

JakeSCahill:
Extracts the `== Metadata` section of connector descriptions into regenerated per-connector partials, so metadata documented upstream flows into published pages automatically instead of being frozen into hand-maintained main pages.

## How it works

- When a connector's description contains a `== Metadata` section (52 of the 65 metadata-documenting components already follow this convention as of Connect 4.99.0), the generator writes it to `modules/components/partials/metadata/<type>/<name>.adoc`, regenerated on every run like the fields and examples partials.
- The intro template replaces the inline metadata block with an `include::` directive pointing at the partial, preserving its position relative to surrounding sections (such as `== Permissions`).
- `migrate-metadata-to-partials.js` is a one-time migration helper for existing pages.
- New `--template-metadata` CLI option to override the partial template.

## Testing

14 new unit tests covering section extraction (edge cases: no section, section at end, empty description), the template banner, include rewriting, and generator output routing. Full suite: 717/718 pass; the one failure is the pre-existing environment-dependent `generate_rpk_docs` MCP integration test, untouched by this change.

## Also included

The rpk download retry fix originally bundled from the closed duplicate #212 now lives in its own PR: #209 (where it belongs — it also carries the review fixes). The port into this branch was reverted.


Feediver1 (commented):
## PR Review: extract connector metadata sections into regenerated partials (#210)

**Scope:** tooling review (extraction logic, tests, migration safety, consumer impact) — not an .adoc content review.
**Overall assessment:** Well-designed and well-tested — the extraction logic is careful (block-delimiter-aware, offset round-trip tested), the migration is safe by design (dry-run default, never clobbers generated content), and all 16 new tests pass locally. One real lifecycle gap (CodeRabbit's stale-partial finding) deserves fixing before this ships, with a better remedy than CodeRabbit proposed.

### Verification performed

- **All 16 new tests pass** (`npx jest __tests__/tools/metadata-partials.test.js`), including the patch-4 hardening cases (headings inside `----` blocks, both as section start and terminator).
- **Full suite: 719/720 pass.** The 1 failure is `__tests__/mcp/integration.test.js:133` — a file this PR doesn't touch; environmental or pre-existing.
- The patch-4 `typeDir` fix is correctly scoped (computed at loop top) and the generator-level test covers path/include-directive agreement.
- Offset arithmetic in `locateMetadata` round-trips exactly (tested); the migration correctly prefers generator-produced partials over page-seeded ones and skips non-connector pages directly under `pages/`.
- Version 5.1.3 → 5.2.0 is appropriate semver; lockfile consistent; CLI_REFERENCE auto-updated.

### Must fix: stale-partial lifecycle (CodeRabbit finding #1, remedy corrected)

When a connector's upstream description *loses* its `== Metadata` section, the generator writes nothing — but the previously generated partial stays on disk, and migrated pages keep including stale metadata forever. CodeRabbit suggests deleting the file, but that's worse: the page's hardcoded include would then be an **unresolved include directive** (broken render + build warning).

**Suggested fix:** when the partial file exists and the new output is empty, overwrite it with an *empty* (or banner-only) file — the include resolves and renders nothing — and log it so a human knows to remove the include line from the page. Add the regeneration test CodeRabbit suggested for this transition.

### Suggestions

1. **Path-convention drift**: fields and examples partials still write under the raw data key (`type`, lines 399/409) while metadata now writes under `typeDir`. Today they coincide, but if they ever diverge, fields/examples inherit the exact mismatch patch 4 fixed for metadata. Consider aligning all three on `typeDir` in a follow-up.
2. **CodeRabbit #3 (CLI registration) has practical merit beyond convention**: consumers install this package via npm, so `node <path-in-node_modules>/migrate-…` is awkward. A registered flag/command would be discoverable — or plan to delete the script after the migration runs.
3. **CodeRabbit #2 (minimatch)**: the 10-line recursive walk is correct; adopt minimatch only if the script becomes long-lived (folds into #2 above).

### Consumer-repo checklist (worth adding to the PR body)

Merging + publishing 5.2.0 does nothing by itself: the migration must be run with `--write` in **rp-connect-docs**, and that repo's regen workflow picks up the new partials on its next run. cloud-docs inherits via the shared source — eyeball the first post-migration sync.

### What works well

- The convention is documented in-code with falsifiable data ("52 of the 65 metadata-documenting components already follow this heading convention as of Connect 4.99.0").
- Patch 4 shows the right instinct: edge cases (delimited blocks, path/include agreement, non-connector pages) found and fixed *with tests* before review caught them.
- The migration's design — dry-run default, generator output preferred over page-seeded content, idempotent re-runs — makes the one-time operation low-risk.
- Test coverage spans unit, template, and end-to-end generator levels, including negative cases.

Feediver1:
Follow-up on the full-suite result in the review above: the `__tests__/mcp/integration.test.js:133` failure is now **confirmed pre-existing**, not a regression from this PR — it fails identically on the #211 branch (`fix/rpk-examples-multiline`), which doesn't touch the metadata code, and neither branch touches any `mcp/` files. No need to investigate it as part of this PR; it likely deserves its own issue (the assertion expects the MCP error message to not contain 'required').

Feediver1 (approved):
Check out CodeRabbit's suggestions

PR #211, opened 2026-07-16
fix(rpk-docs): preserve continuation indent in examples; add k8s plugin

JakeSCahill:
Two fixes for `rpk k8s` docs generation (context: redpanda-data/docs#1801, where these pages were generated manually):

## formatExamples: preserve continuation indentation

Multi-line example commands with backslash continuations were already grouped into a single code block, but each line was trimmed, flattening the conventional continuation indent:

```
rpk k8s multicluster bootstrap \
--kubeconfig /path/to/kubeconfig \
--namespace redpanda
```

Now the run is dedented by its minimum indent instead, preserving relative indentation:

```
rpk k8s multicluster bootstrap \
  --kubeconfig /path/to/kubeconfig \
  --namespace redpanda
```

## KNOWN_PLUGINS: add k8s

The pipeline never installed the Redpanda Kubernetes plugin, so `rpk k8s multicluster` commands could not appear in generated docs. During the 26.2 beta the install step logs a non-fatal warning (`rpk k8s install` finds no `latest` release because the plugin publisher's `stableVersionRe` only promotes pure `X.Y.Z` versions); it becomes effective at GA.

Note: the rpk plugin protocol (`--help-autocomplete`) exposes no flag data, so plugin subcommand flags tables still need the tree augmented from the plugin's own help output. That gap is documented in redpanda-data/docs#1801 and is not addressed here.

## Testing

Two new tests: continuation runs stay in one block with relative indent preserved; blank-line-delimited examples split into distinct blocks. Full rpk-docs suite: 289/289 pass. Version bumped to 5.1.3 with lockfile refresh.


Feediver1 (commented):
## PR Review: preserve continuation indent in examples; add k8s plugin (#211)

**Scope:** tooling review (generator logic, tests, release coordination) — not an .adoc content review.
**Overall assessment:** Both fixes are correct, tested, and honestly scoped — probed the two plausible failure modes in the new code and both are safe. The only real issue is release coordination with #210, which edits the same version lines.

### Verification performed

- **Tests: 39/39 pass** in the touched suite (including the two new cases: continuation-indent preservation, blank-line-delimited block separation). **Full suite 705/706** — the one failure is the same `mcp/integration.test.js` case that also fails on #210's branch, in a file neither PR touches: pre-existing/environmental, not a regression.
- **Probed the empty-`run` hazard**: if an indented `#` comment could reach the dedent branch, `run` would stay empty (`Math.min()` → Infinity, `i` never advances → infinite loop). It can't — the comment branch (`isIndented && startsWith('#')`) is checked *first*, so the while loop always consumes at least the current line. The `#` guard inside the while correctly handles only subsequent lines.
- **Probed the plugin-install failure path**: a failed `rpk k8s install` hits the `console.warn` branch and the loop continues — generation is unaffected. The PR body's beta-window disclosure matches the code exactly.
- Dedent semantics verified against test expectations: relative indentation preserved, comments still lifted out as descriptions, blank lines still split blocks.

### Release coordination (the one must-plan item)

**Version collision with #210.** This PR bumps 5.1.3 → 5.1.4 (after already dodging one collision with #209). #210, also open, bumps 5.1.3 → 5.2.0. Both edit the same `package.json`/`package-lock.json` lines, so whichever merges second needs a manual rebump — git prevents a silent downgrade, but plan the order deliberately. Natural sequence: merge this one (5.1.4, small fix) first, then rebase #210 onto it as 5.2.0.

### Suggestions

1. During the beta window, rpk regen runs will silently produce no `rpk k8s` pages (plugin not installed → commands absent from the tree). Since docs#1681 ships those pages with autogeneration banners, a pre-GA regen could be misread as "the generator skips them." Consider promoting the PR body's beta-window note to a code comment in the warn branch.
2. The `--help-autocomplete` limitation (plugin subcommand flag tables need the tree augmented from the plugin's own help output) is a real known gap worth tracking as an issue rather than a PR-body footnote, so it isn't lost after merge.
3. **Post-merge check**: the first 26.2 `gen-rpk` regen after this merges should be diffed against docs#1681's hand-added `rpk k8s` pages — that diff is the actual end-to-end test of this PR. Expect a broad-but-cosmetic indentation diff across all rpk pages with multi-line examples.

### What works well

- The empty-run hazard turned out to be structurally impossible — branch ordering is right, and the redundant `#` guard is defense-in-depth rather than a bug.
- Honest disclosure of the beta-window warning and the autocomplete limitation — both checked out against the code.
- The version-collision dodge in the second commit shows the right release-number discipline; it just needs one more application against #210.

Feediver1 (approved):
Note the order in which these PRs must be merged. I suspect you are already aware. ;-)

PR #214, opened 2026-07-20
fix(property-docs): stamp Introduced-in versions before AsciiDoc generation

JakeSCahill:
## Problem

Properties added in the release being generated never got their `*Introduced in <tag>*` line in that release's auto-docs PR — it only appeared one release later. Observed in redpanda-data/docs#1806: the three properties new in v26.1.13 (`datalake_coordinator_max_files_per_commit`, `datalake_coordinator_max_pending_files`, `enable_development_metrics`) have `"version": "v26.1.13"` stamped into `property-overrides.json` by that same PR, yet the generated `cluster-properties.adoc` and the `redpanda-properties-v26.1.13.json` attachment carry no version for them. All 56 overrides stamped in earlier releases render fine.

## Root cause

Order of operations in `doc-tools generate property-docs` (`bin/doc-tools.js`):

1. **Phase 1 (extraction)** bakes the overrides file into `redpanda-properties-<tag>.json` — before new properties have a stamp.
2. **Phase 2 (diff)** stamps `version: <newTag>` for newly detected properties via `updatePropertyOverridesWithVersion`, but only into the overrides **file**, never into the already-extracted JSON.
3. **Phase 3 (render)** generates AsciiDoc from the Phase-1 JSON, so `{{#if version}}` sees nothing.

## Fix

New `updatePropertiesJsonWithVersion` in `cli-utils/diff-utils.js`, called immediately after the overrides stamp in Phase 2, patches the same version stamps into the extracted JSON — so the same run's AsciiDoc and JSON attachment both carry `Introduced in <newTag>`.

The stamp is a **surgical textual insertion**, not a parse/re-serialize round-trip: `JSON.stringify` would corrupt values JavaScript numbers cannot represent (uint64 maxima like `18446744073709551615` → `18446744073709552000`), reformat floats (`0.0` → `0`), and reindent the whole Python-generated file. The edited content is validated with `JSON.parse` before writing.

## Testing

- 9 unit tests in `__tests__/cli-utils/diff-utils.test.js`, including byte-for-byte preservation of uint64/float values and a definitions-section collision guard.
- Full `__tests__/tools/` + `__tests__/cli-utils/` suites: 493/493 pass.
- Validated end-to-end against the real redpanda-data/docs#1806 branch data: stamping + `generateAllDocs` produces exactly +3 JSON lines and the three expected `*Introduced in v26.1.13*` blocks in `cluster-properties.adoc`, with zero unrelated churn.

Version bumped to 5.1.5 (5.1.4 is claimed by #211).


Feediver1 (commented):
## Code review (docs-team-standards, tooling focus)

**Overall:** Correct fix with an accurate three-phase root-cause analysis, a well-justified surgical-edit design, and a genuinely strong test suite. Nothing blocking except one trivial release-hygiene error (the version number); three hardening suggestions.

### Verified correct

- **The path assumption holds:** the call site's `path.resolve(outputDir, 'attachments', ...)` matches the extractor's own `OUTPUT_JSON_DIR` (line 923) and two pre-existing uses of the identical resolution (lines 989, 1009).
- **The surgical-edit rationale is real, and proven:** a `JSON.parse`/`stringify` round-trip would corrupt uint64 maxima (`18446744073709551615`) and reformat floats (`0.0 → 0`) across the whole Python-generated file — exactly the "unrelated drift in regen diffs" failure mode the docs team currently hand-verifies against. The byte-preservation test pins this down, including the no-trailing-newline detail.
- **Corruption can't reach disk:** `JSON.parse(raw)` gates the write, so a bad insertion degrades to a warning + unstamped file (the pre-PR behavior), never a broken attachment.
- Idempotence, ghost properties, flat JSON, missing file, and the definitions-name-collision case are all tested; CI is green across Node 18/20 and the Python extractor matrix.

### Must fix (trivial)

1. **Version bump skips a release: `5.1.3 → 5.1.5`.** npm has no published `5.1.4` (latest is 5.1.3), so the gap looks like a typo. Bump to `5.1.4`, or note why 5.1.4 is skipped.

### Suggestions

1. **Bound the key search to the end of the `properties` section.** The anchor prevents false matches *before* `"properties": {` (tested), but not after the section closes — a same-named key in a section that *follows* properties could be stamped if the property were somehow absent from the section. Cheap hardening: cap the search at the section's closing brace, or add the definitions-after-properties variant as a test.
2. **Log silently skipped properties.** `insertVersionIntoRawJson` returns null when the key line doesn't match the expected `"name": {` end-of-line format. If the extractor's formatting ever drifts, stamping quietly no-ops and the original one-release-late bug returns with no signal. When `stampedCount < newProperties.length`, warn with the names that weren't stamped.
3. **Design note:** the JSON stamp runs only inside the `if (overridesPath && exists)` guard, but it depends only on `diffData` — runs without an overrides file still won't get Introduced-in lines. That preserves parity with existing behavior, so fine as-is; worth a one-line comment or a follow-up issue so the coupling is deliberate rather than accidental.

### What works well

- The PR body is a model bug report: symptom with a concrete reproducer PR (docs#1806), ordered root cause, and the fix's design constraint explained
- Test-first coverage of the scary parts (number fidelity, idempotence, collision) rather than just the happy path
- This fix directly de-noises future property-docs regen PRs — the same class of diff-drift the docs team currently has to verify by hand

_🤖 Automated review via docs-team-standards pr-review (tooling focus)._

Feediver1 (approved):
Note the must-fix issue that Claude found

PR #216, opened 2026-07-21
fix: render connector sticky-bar metadata server-side, keep license notice visible

JakeSCahill:
## Problem

Connector pages flash their metadata on load, and the enterprise license notice disappears entirely ([DOC-2383](https://redpandadata.atlassian.net/browse/DOC-2383)): the `component_type_dropdown` macro emits a `.metadata-block` in the body, and the docs-ui client-side script moves the Type dropdown and availability info into the sticky bar after `DOMContentLoaded`, hiding the whole block - including the license notice it never moved.

The macro already tried to set the sticky-bar page attributes that the UI renders server-side (`article.hbs`), but they can never reach UI templates from an Asciidoctor extension: Antora extracts page attributes in a **header-only parse with extensions disabled**, and `convertDocument` only re-extracts metadata when `page.asciidoc` is unset - which it never is, because `convertDocuments` always sets it in that header-only pass. The availability attributes were also missing the `page-` prefix that Antora requires to expose them to templates.

## Changes

- **`extensions/generate-rp-connect-info.js`**: new `documentsConverted` handler sets `page-context-switcher`, `page-cloud-available(-url)`, `page-self-managed-available(-url)`, and `page-self-managed-only` directly on `page.asciidoc.attributes`, matching translated CSV rows to pages by `pub.url`. Mutating page metadata after conversion is the only way computed attributes reach UI templates.
- **`macros/rp-connect-components.js`**: `component_type_dropdown` no longer emits the in-body Type dropdown, "Available in:" paragraph, or inline dropdown script (all now rendered server-side by the UI from the attributes above). It emits only the license notice, which stays visible in the body. The dead attribute-setting code is removed with a comment explaining why it can never work.

## Testing

Verified end-to-end in an isolated Antora build using this branch, a CSV fixture, and the current UI bundle plus the companion docs-ui change (redpanda-data/docs-ui#403):

- Sticky-bar metadata renders at build time (no flash): context switcher with the current type first and correct per-variant links, "Also in Cloud" / "Also in Self-Managed" / "Self-Managed Only" badges all correct.
- License notice stays visible in the body for enterprise connectors.
- No duplicated in-body metadata; the docs-ui script early-returns when the server-rendered sticky bar is present, and remains a fallback for content built with older extension versions.

## Rollout

Release the docs-ui change first (redpanda-data/docs-ui#403 - fixes the disappearing license notice on its own), then this package, then rebuild the docs to eliminate the flash entirely.

[DOC-2383]: https://redpandadata.atlassian.net/browse/DOC-2383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/extensions/generate-rp-connect-info.js
+++ b/extensions/generate-rp-connect-info.js
@@ -30,9 +30,23 @@ module.exports.register = function ({ config }) {
     return require('../cli-utils/octokit-client')
   }
 
+  // Translated CSV rows, kept for the documentsConverted hook below
+  let translatedRows = null
+
   // Use 'on' and return the promise so Antora waits for async completion
   this.on('contentClassified', ({ contentCatalog }) => {
     return processContent(contentCatalog)
+  })
+
+  // Set sticky-bar page attributes (type context switcher and availability badges) so the UI
+  // templates (article.hbs) render them server-side, with no flash of client-side rearranging.
+  // This must run on documentsConverted: Antora extracts page attributes for UI templates in a
+  // header-only parse with Asciidoctor extensions disabled, so these attributes cannot be set
+  // from the component_type_dropdown macro during conversion. Mutating page.asciidoc.attributes
+  // after conversion is the supported way to feed computed attributes to UI templates.
+  this.on('documentsConverted', ({ contentCatalog }) => {
+    if (!translatedRows) return
+    setStickyBarPageAttributes(contentCatalog, translatedRows, logger)
   })
 
   async function processContent (contentCatalog) {
@@ -56,6 +70,7 @@ module.exports.register = function ({ config }) {
       const parsedData = Papa.parse(csvData, { header: true, skipEmptyLines: true })
       const enrichedData = translateCsvData(parsedData, pages, logger)
       parsedData.data = enrichedData
+      translatedRows = enrichedData
 
       // Set csvData on all relevant components
       const componentsToEnrich = [redpandaConnect, redpandaCloud, preview].filter(Boolean)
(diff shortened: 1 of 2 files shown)


micheleRP (approved):
Approving. Traced the two things this fix hinges on and both hold up, and CI (test-extensions / test-macros on Node 18 + 20) is green.

**Verified**
- URL matching is safe by construction: `row.redpandaConnectUrl`/`redpandaCloudUrl` are set to `connectPage.pub.url`/`cloudPage.pub.url` in `translateCsvData`, and `setStickyBarPageAttributes` keys its lookup map on `page.pub.url` — same value, same source, no path-format mismatch. Unmatched pages fall back to the client-side script, so the failure mode is graceful.
- Removing the inline dropdown `<script>` is safe: the second definition of `window.toggleComponentTypeDropdown` (line ~747, catalog/table macro, guarded by `if (!toggle || !menu) return`) is independent and untouched, and connector pages no longer emit any button that calls the removed copy.

**Suggestions (non-blocking, follow-up)**
1. No regression coverage for the new logic. `setStickyBarPageAttributes` and `buildContextSwitcher` are the heart of the fix but aren't exercised by the passing `test-macros` job. A small test over a multi-type / enterprise / cloud-only fixture asserting the attributes set would lock the behavior in. Fine as a follow-up.
2. Minor robustness: `buildContextSwitcher` calls `row.type.trim()` unguarded — a CSV row missing `type` would throw. Very low risk, but `(row.type || '').trim()` would match the guarding style already used for `row.connector`.

Note: this deliberately diverges from the ticket's suggested `setHeaderAttribute()` approach — the `documentsConverted` mutation of `page.asciidoc.attributes` is the mechanism that actually reaches UI templates, and the inline comment explaining why is a good addition.

PR #218, opened 2026-07-22
feat(connect): normalize metadata partial field-list formatting

JakeSCahill:
## Problem

Metadata partials are rendered **verbatim** from each connector's upstream `== Metadata` section, so formatting is inconsistent across connectors:

- some wrap the field list in a fenced block of bare names (```` ```text ```` / ```` ``` ````, e.g. `http_server`, `csv`, `gcp_cloud_storage`, `nats_kv`)
- others use an AsciiDoc bullet list with the field name in inline code (e.g. `oracledb_cdc`)

Reported on redpanda-data/rp-connect-docs#459.

## Fix

Add `normalize-metadata.js` and apply it in `renderConnectMetadata`. For an extracted metadata block it:

- **strips the fences** around a bare field list and renders the fields as a plain bullet list;
- **inline-codes the field name** in each field bullet (`` - `http_server_user_agent` ``), keeping annotations (`` - `mod_time` (RFC3339)``) and descriptions (`` - `operation`: … ``);
- leaves **descriptive bullets** ("All headers (only first values are taken)") as prose;
- leaves **already inline-coded** bullets untouched;
- **preserves `===` subheadings** that legitimately group metadata by operation (e.g. the `nats_kv` processor);
- keeps **non-field fenced blocks** (YAML examples) verbatim.

Field names are detected as lowercase snake_case leading tokens, which cleanly distinguishes them from descriptive bullets.

## Tests

`npx jest __tests__/tools/metadata-partials.test.js` → 23 passing, including a new `normalize-metadata` suite covering the ```` ```text ````/bare-fence/nats_kv/`mod_time (RFC3339)`/YAML-example cases. Two existing assertions updated to expect inline-coded field names.

## Related
- The "extra content" side of #459 (mis-scoped `===` under Metadata) is fixed at the source: redpanda-data/benthos#462 (csv), redpanda-data/connect#4620 (gcp_cloud_storage).
- Complements the block-terminator fix in #217.


Feediver1:
Recommended fix:
AsciiDoc listing blocks aren't guarded. extractMetadata (in metadata-utils.js) explicitly tracks ---- delimiters so it doesn't misread headings inside them — evidence that ---- blocks occur in these descriptions. But normalizeMetadataBlock only recognizes ``` and ~~~. If a connector's Metadata section wraps content in ----, any - field_name line inside that literal block gets backtick-wrapped, and backticks inside a literal block render as visible backticks on the page — a regression the current tests wouldn't catch. Fix is small: track BLOCK_DELIMITER =  /^-{4,}$/ state in the main loop and pass those regions through verbatim, plus one test.

Minor suggestions:
- Mismatched fence close: FENCE_CLOSE matches either ``` or ~~~, so a block opened with ``` can be "closed" by ~~~. Cheap fix: capture the opening fence and require the same one to close.
- False-positive surface in FIELD_BULLET: any bullet starting with a lowercase word followed by :  or - gets coded (for example a hypothetical - note: this only applies to... would become - `note`: ...). The heuristic is reasonable and the PR body owns it; just be aware the first regeneration over all connectors should be diff-reviewed rather than trusted (consistent with how we handle property regen drift).
- Unclosed-fence handling is correct (falls back to verbatim) but untested — a one-liner test would pin it.

Feediver1 (approved):

JakeSCahill:
Thanks @Feediver1 — addressed all three in `1494be0a`:

1. **`----` literal blocks** (recommended fix): `normalizeMetadataBlock` now tracks `BLOCK_DELIMITER` (`/^-{4,}$/`) and passes `----` … `----` regions through verbatim, so field-name lines inside a listing block are never inline-coded (backticks would render literally there).
2. **Mismatched fence close**: a fenced block now only closes on the *same* marker it opened with, so a ```` ``` ````-opened block isn't closed by `~~~`.
3. **CodeRabbit nitpick / your false-positive note**: `isFieldListFence` now requires at least one real field-name bullet (`FIELD_BULLET`) before stripping fences, so a fenced block of purely descriptive bullets keeps its fences.

Left the `FIELD_BULLET` lowercase-word heuristic as-is per your note (the PR body owns it), and added regression tests for all three.

**End-to-end verified:** ran the generation helper over all **52** real 4.102.0 connectors with a `== Metadata` section — 0 residual field-list fences, 0 uncoded field bullets, 0 anomalies; `===` grouping preserved (csv, nats_kv). Full template chain (`metadata-partials.hbs`) renders `amqp_1` correctly. Unit suite 31/31.

(No current connector has a `----` block inside its Metadata section, so fix #1 is a defensive guard.)

PR #220, opened 2026-07-24
fix(connect): render map fields as object in field reference

JakeSCahill:
## Problem

In the connector field reference, string-valued **map** fields render `*Type*: string`, which is misleading — they accept a map/object, not a string. Examples: `redpanda_migrator.headers` and `oracledb_cdc.snapshot_filters` (both `type: string, kind: map`). Flagged during review of rp-connect-docs#466.

## Cause

`renderConnectFields` normalized types with type-restricted cases:

```js
} else if (child.type === 'unknown' && child.kind === 'map') {
  displayType = 'object';
} else if (child.type === 'string' && child.kind === 'array') {
  displayType = 'array';
...
} else {
  displayType = child.type;   // string map lands here -> "string"
}
```

Only `type: unknown` maps became `object`; a `type: string, kind: map` field fell through to `child.type` (`string`).

## Fix

Detect maps and lists/arrays by **`kind`**, regardless of the value type:

```js
} else if (child.kind === 'map') {
  displayType = 'object';
} else if (child.kind === 'array' || child.kind === 'list') {
  displayType = 'array';
} else {
  displayType = child.type;
}
```

This also generalizes array handling (e.g. an `int` array now renders `array` instead of `int`, consistent with how string arrays already rendered).

## Tests

`npx jest __tests__/tools/generate-rpcn-connector-docs.test.js` → 25 passing, including two new cases: a string-valued map renders as `object` (not `string`), and a non-string array renders as `array`.

## Release / merge order

Bumped to **5.2.5**. #218 is still open at **5.2.4** — please merge **#218 before this PR** so both publish (publish is skipped if a version already exists on npm). If this merges first, #218 must rebase and re-bump.

### DIFF
--- a/__tests__/tools/generate-rpcn-connector-docs.test.js
+++ b/__tests__/tools/generate-rpcn-connector-docs.test.js
@@ -429,6 +429,24 @@ describe('Utility Helpers', () => {
       expect(result).toContain("*Default*: `30`");
     });
 
+    it('renders a string-valued map field as object, not string', () => {
+      const mapField = [
+        { name: 'headers', type: 'string', kind: 'map', description: 'Custom headers keyed by name.' },
+      ];
+      const result = renderConnectFields(mapField).toString();
+      expect(result).toContain('=== `headers`');
+      expect(result).toContain('*Type*: `object`');
+      expect(result).not.toContain('*Type*: `string`');
+    });
+
+    it('renders any list/array field as array regardless of element type', () => {
+      const listField = [
+        { name: 'ports', type: 'int', kind: 'array', description: 'Ports.' },
+      ];
+      const result = renderConnectFields(listField).toString();
+      expect(result).toContain('*Type*: `array`');
+    });
+
     it('recursively renders nested children with correct path', () => {
       const result = renderConnectFields(nestedField).toString();
       expect(result).toContain("=== `parent`");
--- a/__tests__/tools/generate-rpcn-connector-docs.test.js
+++ b/__tests__/tools/generate-rpcn-connector-docs.test.js
@@ -439,13 +439,17 @@ describe('Utility Helpers', () => {
       expect(result).not.toContain('*Type*: `string`');
     });
 
-    it('renders any list/array field as array regardless of element type', () => {
-      const listField = [
-        { name: 'ports', type: 'int', kind: 'array', description: 'Ports.' },
-      ];
-      const result = renderConnectFields(listField).toString();
-      expect(result).toContain('*Type*: `array`');
-    });
+    it.each(['array', 'list', '2darray'])(
+      'renders a field of kind %s as array regardless of element type',
+      (kind) => {
+        const listField = [
+          { name: 'ports', type: 'int', kind, description: 'Ports.' },
+        ];
+        const result = renderConnectFields(listField).toString();
+        expect(result).toContain('*Type*: `array`');
+        expect(result).not.toContain('*Type*: `int`');
+      }
+    );
 
     it('recursively renders nested children with correct path', () => {
       const result = renderConnectFields(nestedField).toString();
--- a/__tests__/tools/generate-rpcn-connector-docs.test.js
+++ b/__tests__/tools/generate-rpcn-connector-docs.test.js
@@ -483,6 +483,17 @@ describe('Utility Helpers', () => {
       expect(result).not.toContain('array<');
     });
 
+    it('falls back to plain array when the element type is itself array (override-forced types)', () => {
+      // docs-data/overrides.json in rp-connect-docs sets `type: "array"` on
+      // some fields (for example, NATS `urls`) to force a plain-array display.
+      const listField = [
+        { name: 'urls', type: 'array', kind: 'array', description: 'URLs.' },
+      ];
+      const result = renderConnectFields(listField).toString();
+      expect(result).toContain('*Type*: `array`');
+      expect(result).not.toContain('array<');
+    });
+
     it('recursively renders nested children with correct path', () => {
       const result = renderConnectFields(nestedField).toString();
       expect(result).toContain("=== `parent`");
--- a/__tests__/tools/generate-rpcn-connector-docs.test.js
+++ b/__tests__/tools/generate-rpcn-connector-docs.test.js
@@ -429,27 +429,59 @@ describe('Utility Helpers', () => {
       expect(result).toContain("*Default*: `30`");
     });
 
-    it('renders a string-valued map field as object, not string', () => {
-      const mapField = [
-        { name: 'headers', type: 'string', kind: 'map', description: 'Custom headers keyed by name.' },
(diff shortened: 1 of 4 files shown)


Feediver1:
## Review: render map fields as `object`

### Overview
Fixes the connector field reference so map-kind fields render `*Type*: object` regardless of value type (previously only `type: unknown` maps did, so `type: string, kind: map` fields like `redpanda_migrator.headers` rendered as `string`). Also generalizes array/list handling to match on `kind` alone. Includes two new tests and a version bump to 5.2.5.

### Correctness — verified
- The refactor is sound: the `kind` checks sit after the `isArrayTitle` special case and before the `child.type` fallback, so precedence is preserved. Scalar object fields (`type: object, kind: scalar`) still render `object` via the fallback, and object-valued maps render `object` as before.
- The consolidation from four type-restricted branches to two kind-based ones changes exactly the cases the PR intends: string/typed maps → `object`, typed arrays → `array`. The `[]` name-suffix logic (`kind === 'array'`) is untouched, so headings are unchanged.
- All CI checks pass (Node 18/20 test matrices, MCP integration, CLI contract).

### Release ordering — now resolved
The PR body asks that #218 merge first. That happened: #218 merged this morning and 5.2.4 published to npm shortly after. So the 5.2.5 bump here is unblocked and correct, the lockfile is consistent, and the PR is mergeable as-is. The merge-order warning in the body is now stale — worth a one-line edit so a reviewer doesn't hold the PR waiting on it.

### Suggestions (all minor)
- **Object arrays lose element info.** `type: object, kind: array` previously rendered `object`, now `array` — both are lossy, but the codebase already has an `array<object>` display for the `isArrayTitle` case. Rendering `array<object>` (or `array<int>` etc.) when the element type is known would be strictly more informative. Fine as a follow-up; the current change is still a net improvement.
- **`2darray` kind is still unhandled** (Benthos field specs have it, e.g. `workflow.order`). It falls through to `child.type` and renders the element type. Pre-existing gap, not a regression — noting it since this PR is already in the type-normalization business.
- **Test title vs. coverage**: "renders any list/array field as array" only exercises `kind: 'array'`. Since `kind: 'list'` handling is part of the changed branch, a second one-liner with `kind: 'list'` would pin it.
- **Comment style**: the block comment narrates the old behavior ("Previously only `type: unknown` maps were handled…"). That history lives in the commit/PR; a single constraint-style line ("maps always render as `object`, regardless of value type") ages better.

### Risks
Low. Display-only change in doc generation; no security or performance surface. The one thing to double-check downstream is churn in generated connector docs — every typed-array field flips from its element type to `array`, which will touch many pages in rp-connect-docs on the next regeneration. That's the intended fix, but the regeneration diff will be broad, not limited to the two motivating fields.

**Verdict**: correct, well-tested, and cleanly scoped — good to merge; suggestions are all optional polish.

Feediver1 (approved):
The suggestions are your call.

JakeSCahill:
Thanks for the review @Feediver1. Follow-ups pushed in 5e210f0 — breakdown of each suggestion:

**Applied**

- **`2darray` kind handling**: verified against the real docs data — `workflow.order` is `kind: 2darray, type: string` and fell through to the type fallback, rendering `*Type*: string`. That's the same bug class this PR fixes, so `2darray` now joins the kind-based array branch and renders `array`. (`processors.workflow.input_batches`/`output_batches` are also `2darray` and flip from `object` to `array`.)
- **Test coverage for `kind: 'list'`** (also CodeRabbit's actionable comment): the array test is now parameterized with `it.each(['array', 'list', '2darray'])`, with an added negative assertion that the element type doesn't leak through.
- **Comment style**: since the branch changed anyway, the block comments now state the constraint ("maps always render as `object`…") instead of narrating the old behavior.

**Declined (deferred as follow-up)**

- **`array<element>` display for typed arrays**: agreed it would be strictly more informative, but as you noted it's follow-up material — it changes the rendering contract this PR deliberately sets (`array` regardless of element type) and would broaden the regeneration diff further. Happy to pick it up separately.

Validation: `node --check` clean on both changed files; full suite passes (33 suites, 754 tests).

JakeSCahill:
Update: the `array<element>` display suggestion is no longer deferred. Reversing the "declined" note in my previous comment — per offline discussion, it lands in this PR after all. Implemented in 1bfe5a2.

**What the display looks like now**

| Field spec | Before (5e210f0) | After |
|---|---|---|
| `kind: array, type: string` (for example `workflow.branch_resources`) | `array` | `array<string>` |
| `kind: array, type: int` | `array` | `array<int>` |
| `kind: array, type: processor` (for example `pipeline.processors`) | `array` | `array<processor>` |
| `kind: 2darray, type: string` (for example `workflow.order`) | `array` | `array<array<string>>` |
| `kind: array, type: unknown` or no type | `array` | `array` (fallback) |
| `kind: map` (any value type) | `object` | `object` (unchanged) |

Element types come straight from the field spec's `type`, which the docs data always carries (`unknown` when untyped). This matches the existing `array<object>` display that `[]`-suffixed object sections already used, so the field reference is now internally consistent. Maps stay `object` per this PR's earlier decision — surfacing map value types is a separate conversation if we ever want it.

**Regeneration impact**: broader than before, as expected. Against the 4.53.0 docs data, roughly 350 array fields flip from plain `array` to a typed display (including component-typed ones such as `array<processor>`), plus the three `2darray` fields. One `array/unknown` field keeps the plain `array` fallback.

Validation: full suite green (33 suites, 763 tests, up from 754 with the new parameterized cases).

JakeSCahill:
Ran a full end-to-end validation of this branch: the real `doc-tools generate rpcn-connector-docs` pipeline (same command as the auto-docs workflow) against the checked-in RPCN 4.102.0 data, followed by two full Antora site builds (baseline vs this branch).

**One real bug found and fixed** (`e4401a0`): the production `overrides.json` in rp-connect-docs pins `type: "array"` on 21 array-kind fields (NATS-family `urls` etc.), and the typed-array display rendered those as the degenerate `array<array>`. Element type `array` is now treated as unknown and falls back to plain `array`, with a regression test. Full suite: 764/764 pass.

**E2E results:**
- 444 type-line flips across 191 field partials, and nothing else changes — every diff line is a `*Type*:` line. Breakdown: 202 typed scalars (`array<string>`, `array<int>`, ...), 117 `object` → `array<object>` (previously mislabeled), 76 `array<processor>`, 48 maps → `object`, 5 component-typed arrays, 1 `array<array<string>>` (`workflow.order`).
- Antora build logs baseline vs branch: byte-identical (zero new warnings or errors).
- Rendered HTML: angle brackets escape correctly as literal text inside monospace on all 173 affected pages, no swallowed tags.
- Whats-new tables are unaffected (they carry component types, not field types).

Follow-up candidate (separate PR): the 21 `type: "array"` overrides in rp-connect-docs `overrides.json` are now redundant and can be removed.

PR #222, opened 2026-07-28
fix(cli-utils): map cross-component doc URLs to qualified xrefs in urlToXref

JakeSCahill:
## Bug

`urlToXref` in `cli-utils/convert-doc-links.js` converts docs.redpanda.com URLs found in operator godoc (used by `doc-tools generate crd-docs` and `generate helm-docs`) into Antora xrefs by unconditionally treating the first URL path segment as a **module in the current component**. For URLs that point at other Antora components on docs.redpanda.com, this produces broken xrefs.

Exact reproduction:

| Input | Before (broken) | After |
|---|---|---|
| `https://docs.redpanda.com/redpanda-connect/configuration/secrets/` | `xref:redpanda-connect:configuration/secrets.adoc` | `xref:connect:configuration:secrets.adoc` |

The broken xref shipped in the generated `modules/reference/pages/k-crd.adoc` via redpanda-data/docs#1830 and was hand-fixed there (https://github.com/redpanda-data/docs/pull/1830#issuecomment-5101970808). Every CRD regen reintroduces the breakage until the generator is fixed.

## Fix

Add a verified URL-slug → Antora-component map. When a URL has no legacy `/docs`, `/vX.Y`, or `/current` prefix and its first path segment is in the map, emit a fully qualified `xref:component:module:page.adoc` resource ID with correct colon separation between module and page. A version segment directly after the slug (for example `/streaming/current/...`) is stripped. All other URLs keep the existing same-component behavior, and the map is deliberately not consulted after a legacy prefix is stripped (so `/current/home/...` still resolves to the current component's `home` module).

Shipped mapping (every entry verified against the target repo's `antora.yml` `name:` key; legacy slugs additionally verified via the live site's 301 redirects):

| URL slug | Component | Verification |
|---|---|---|
| `redpanda-connect` (legacy), `connect` | `connect` | rp-connect-docs `antora.yml`; live 301 `/redpanda-connect/...` → `/connect/...` |
| `redpanda-cloud` (legacy), `cloud-data-platform` | `cloud-data-platform` | cloud-docs `antora.yml`; live 301 `/redpanda-cloud/...` → `/cloud-data-platform/...` |
| `redpanda-labs` (legacy), `labs` | `labs` | redpanda-labs `docs/antora.yml`; live 301 `/redpanda-labs/` → `/labs/` |
| `streaming` | `streaming` | redpanda-data/docs `antora.yml`; live 301 `/current/...` → `/streaming/current/...` |
| `agentic-data-plane` | `agentic-data-plane` | adp-docs `antora.yml` |
| `home`, `data-platform`, `self-managed` | same | docs-site umbrella components' `antora.yml` files |

## Tests

New `__tests__/cli-utils/convert-doc-links.test.js` (18 cases), including:

- the exact docs#1830 reproduction pair, with and without trailing slash
- same-component `/current/`, `/docs/`, and `/vX.Y/` URLs unchanged
- mapped URL with a deeper path (`/connect/components/inputs/kafka/` → `xref:connect:components:inputs/kafka.adoc`)
- legacy cloud and labs slugs, versioned `streaming` URLs, module-only and component-only URLs
- label preservation and the legacy-prefix/slug-map precedence rule

Full suite green: 34 suites, 768 tests passed. Also validated the converter against every docs.redpanda.com URL currently present in redpanda-data/redpanda-operator source: all legacy URLs unchanged, and the one cross-component URL now emits the same xref as the manual fix in docs#1830.

### DIFF
--- a/__tests__/cli-utils/convert-doc-links.test.js
+++ b/__tests__/cli-utils/convert-doc-links.test.js
@@ -1,0 +1,121 @@
+'use strict';
+
+const { urlToXref } = require('../../cli-utils/convert-doc-links.js');
+
+describe('urlToXref', () => {
+  describe('validation', () => {
+    it('throws on an invalid URL', () => {
+      expect(() => urlToXref('not-a-url')).toThrow(/Invalid URL/);
+    });
+
+    it('throws on a non-docs.redpanda.com URL', () => {
+      expect(() => urlToXref('https://example.com/current/foo/')).toThrow(
+        /Not a docs\.redpanda\.com URL/
+      );
+    });
+  });
+
+  describe('same-component URLs (unchanged behavior)', () => {
+    it('converts a /current/ URL to a module xref in the current component', () => {
+      expect(
+        urlToXref('https://docs.redpanda.com/current/manage/kubernetes/manage-resources/')
+      ).toBe('xref:manage:kubernetes/manage-resources.adoc');
+    });
+
+    it('converts a legacy /docs/ URL to a module xref in the current component', () => {
+      expect(
+        urlToXref('https://docs.redpanda.com/docs/reference/cluster-properties/')
+      ).toBe('xref:reference:cluster-properties.adoc');
+    });
+
+    it('converts a versioned /vX.Y/ URL to a module xref in the current component', () => {
+      expect(
+        urlToXref('https://docs.redpanda.com/23.2/manage/security/authentication/')
+      ).toBe('xref:manage:security/authentication.adoc');
+    });
+
+    it('converts the site root to the index page', () => {
+      expect(urlToXref('https://docs.redpanda.com/')).toBe('xref:index.adoc');
+    });
+
+    it('preserves a bracketed label', () => {
+      expect(
+        urlToXref('https://docs.redpanda.com/current/manage/kubernetes/manage-resources/[Manage resources]')
+      ).toBe('xref:manage:kubernetes/manage-resources.adoc[Manage resources]');
+    });
+
+    it('does not apply the component map after a legacy prefix is stripped', () => {
+      // /current/home/... is the `home` module of the current component,
+      // not the docs-site `home` umbrella component.
+      expect(
+        urlToXref('https://docs.redpanda.com/current/home/index/')
+      ).toBe('xref:home:index.adoc');
+    });
+  });
+
+  describe('cross-component URLs', () => {
+    it('converts the exact redpanda-connect reproduction case from docs#1830', () => {
+      // Previously produced the broken xref:redpanda-connect:configuration/secrets.adoc
+      expect(
+        urlToXref('https://docs.redpanda.com/redpanda-connect/configuration/secrets/')
+      ).toBe('xref:connect:configuration:secrets.adoc');
+    });
+
+    it('handles the URL without a trailing slash identically', () => {
+      expect(
+        urlToXref('https://docs.redpanda.com/redpanda-connect/configuration/secrets')
+      ).toBe('xref:connect:configuration:secrets.adoc');
+    });
+
+    it('maps the current connect slug the same as the legacy slug', () => {
+      expect(
+        urlToXref('https://docs.redpanda.com/connect/configuration/secrets/')
+      ).toBe('xref:connect:configuration:secrets.adoc');
+    });
+
+    it('handles deeper paths by keeping subdirectories in the page path', () => {
+      expect(
+        urlToXref('https://docs.redpanda.com/connect/components/inputs/kafka/')
+      ).toBe('xref:connect:components:inputs/kafka.adoc');
+    });
+
+    it('maps the legacy redpanda-cloud slug to the cloud-data-platform component', () => {
+      expect(
+        urlToXref('https://docs.redpanda.com/redpanda-cloud/get-started/whats-redpanda-cloud/')
+      ).toBe('xref:cloud-data-platform:get-started:whats-redpanda-cloud.adoc');
+    });
+
+    it('maps the legacy redpanda-labs slug to the labs component', () => {
+      expect(
+        urlToXref('https://docs.redpanda.com/redpanda-labs/docker-compose/single-broker/')
+      ).toBe('xref:labs:docker-compose:single-broker.adoc');
+    });
+
+    it('qualifies streaming URLs and strips the version segment after the slug', () => {
+      expect(
+        urlToXref('https://docs.redpanda.com/streaming/current/manage/kubernetes/manage-resources/')
+      ).toBe('xref:streaming:manage:kubernetes/manage-resources.adoc');
+      expect(
+        urlToXref('https://docs.redpanda.com/streaming/25.1/manage/kubernetes/manage-resources/')
+      ).toBe('xref:streaming:manage:kubernetes/manage-resources.adoc');
+    });
+
+    it('converts a module-only URL to the module index page', () => {
+      expect(
+        urlToXref('https://docs.redpanda.com/connect/configuration/')
+      ).toBe('xref:connect:configuration:index.adoc');
+    });
+
+    it('converts a component-only URL to the component index page', () => {
+      expect(
+        urlToXref('https://docs.redpanda.com/redpanda-connect/')
+      ).toBe('xref:connect::index.adoc');
+    });
+
+    it('preserves a bracketed label on cross-component URLs', () => {
+      expect(
+        urlToXref('https://docs.redpanda.com/redpanda-connect/configuration/secrets/[Secrets]')
+      ).toBe('xref:connect:configuration:secrets.adoc[Secrets]');
+    });
+  });
+});
--- a/__tests__/cli-utils/convert-doc-links.test.js
+++ b/__tests__/cli-utils/convert-doc-links.test.js
@@ -119,3 +119,22 @@ describe('urlToXref', () => {
     });
   });
 });
+
+describe('urlToXref fragment preservation', () => {
+  const { urlToXref } = require('../../cli-utils/convert-doc-links.js');
+
+  test('keeps the anchor on a legacy-prefixed URL', () => {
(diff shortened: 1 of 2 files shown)


JakeSCahill:
Pushed a follow-up commit (499b073) fixing a defect found during the end-to-end test of this PR against the real connect 4.103.0 data: `urlToXref` silently dropped URL fragments, so a real dump URL like `.../reference/properties/cluster-properties/#kafka_batch_max_bytes` converted to an xref that lands at the top of the page instead of the property. The bug pre-dates this branch, but since this PR already reworks `urlToXref`, fixing it here avoids a guaranteed conflict with a separate PR.

Fragments now append to the xref target before the label (`xref:reference:properties/cluster-properties.adoc#kafka_batch_max_bytes`), with three new tests covering the legacy-prefix, cross-component-with-label, and no-fragment cases. Full suite 771/771 green.

JakeSCahill:
CodeRabbit's two nitpicks addressed in d6b446e: added assertions for the `beta` version segment and every previously untested slug-map entry (`agentic-data-plane`, `home`, `data-platform`, `self-managed`) — 24 tests on the module now, full suite 774 green. Skipped the suggested helper extraction: the cross-component branch is 25 lines with the reasoning in comments, and restructuring it mid-review after the fragment fix would add churn without changing behavior.

micheleRP:
Reviewed the logic against the actual `antora.yml` files in each target repo. The approach is right and the fragment-preservation commit is a good catch, but two of the slug-map entries would emit broken xrefs — the same failure class this PR fixes. Worth fixing before merge; nothing here needs a redesign.

## Fix before merge

**1. `labs` is not a component name.**

```js
// redpanda-labs (docs/antora.yml name: labs)
'redpanda-labs': 'labs', // legacy slug
'labs': 'labs',
```

The only `antora.yml` in `redpanda-labs` is `docs/antora.yml`, and its name is **`redpanda-labs`**, not `labs`:

```console
$ find . -name antora.yml -not -path './node_modules/*' | while read f; do printf "%-24s %s\n" "$f" "$(grep -E '^name:' "$f")"; done
./docs/antora.yml        name: redpanda-labs
```

So `https://docs.redpanda.com/redpanda-labs/docker-compose/single-broker/` produces `xref:labs:docker-compose:single-broker.adoc`, pointing at a component that doesn't exist. `redpanda-labs` is aggregated into the site (`antora-playbook.yml:77`), so this is a live broken link. The test at "maps the legacy redpanda-labs slug to the labs component" asserts the wrong value, so the suite currently locks the bug in.

Both slugs should map to `redpanda-labs`.

**2. Component-only URLs assume a ROOT-module `index.adoc` that isn't always there.**

`/redpanda-connect/` → `xref:connect::index.adoc`, which Antora reads as "ROOT module of `connect`". But rp-connect-docs has `start_page: home:index.adoc` and its `modules/ROOT/pages/` is empty — the landing page is `modules/home/pages/index.adoc`. So that xref doesn't resolve. `cloud-data-platform` has the same shape (empty `modules/ROOT/pages/`).

Of the mapped components, only `home`, `self-managed`, and `data-platform` actually have `ROOT/pages/index.adoc`. Resolving component-only URLs via the target's `start_page` would handle all of them.

## Suggestion

**The "first segment after the slug is a module" heuristic misfires on ROOT-module pages.** docs-site sets `html_extension_style: indexify` (`antora-playbook.yml:58`), so every page URL ends in `/` and `/data-platform/overview/` is genuinely ambiguous between "module `overview`, page `index`" and "ROOT page `overview.adoc`".

`data-platform`, `self-managed`, and `home` each have **only** a ROOT module, so for those the second reading is the correct one: `xref:data-platform::overview.adoc`, not the asserted `xref:data-platform:overview:index.adoc`. Same applies to the `self-managed` assertion. Either consult the content catalog, or restrict the module assumption to components known to be multi-module.

## What works well

- The `urlToXref` reproduction case from docs#1830 is fixed correctly, and `connect`, `cloud-data-platform`, `streaming`, and `agentic-data-plane` all check out against their `antora.yml`.
- The fragment-preservation commit is a real find that stands on its own — dropped anchors are invisible because Antora doesn't validate fragments.
- Adding a test per map entry is the right instinct, since a wrong entry fails silently. That's exactly what surfaced the `labs` issue.

JakeSCahill:
Both findings addressed:

- **Component-only URLs**: confirmed — connect and cloud-data-platform (and streaming/agentic-data-plane) have no ROOT index.adoc; their `start_page` is `home:index.adoc`. Component-only URLs now resolve to the component's start page, and the tests assert the corrected values plus the ::index.adoc case for components that do have a ROOT index (labs, umbrella components).
- **labs naming**: I verified rather than assumed — redpanda-labs' `docs/antora.yml` says `name: labs` on main, which is the branch the site playbook builds, and live `/labs/` serves 200 while `/redpanda-labs/` 301s to it. So the map keeps `labs` as the component with `redpanda-labs` as the legacy slug. If you saw a build where labs resolves as `redpanda-labs`, keen to see it, but three signals say labs is canonical.

micheleRP (dismissed):
Solid diagnosis and the reproduction table makes the bug easy to confirm. I independently verified every entry in `COMPONENT_SLUG_MAP` against each repo's `antora.yml` on `main` — all correct:

| Slug | Component | Verified |
|---|---|---|
| `redpanda-connect`, `connect` | `connect` | rp-connect-docs `antora.yml` |
| `redpanda-cloud`, `cloud-data-platform` | `cloud-data-platform` | cloud-docs `antora.yml` |
| `redpanda-labs`, `labs` | `labs` | redpanda-labs `docs/antora.yml` @ main |
| `streaming` | `streaming` | redpanda-data/docs `antora.yml` |
| `agentic-data-plane` | `agentic-data-plane` | adp-docs `antora.yml` |

`COMPONENT_START_PAGE` checks out too — `streaming`, `connect`, `cloud-data-platform`, and `agentic-data-plane` all declare `start_page: home:index.adoc`, and `labs` plus the docs-site umbrellas have a real ROOT `index.adoc`. (My own local redpanda-labs checkout still says `name: redpanda-labs`; it's six months stale, so your `labs` reading is the right one.)

Requesting changes on one correctness issue that the tests currently lock in, plus one design question.

## 1. The three docs-site umbrella components are ROOT-only, so their mappings emit broken xrefs

`docs-site/home`, `docs-site/self-managed`, and `docs-site/data-platform` each contain **only `modules/ROOT/`**. The "first path segment after the slug is a module" rule doesn't hold for a single-module component:

| URL | This PR emits | Correct |
|---|---|---|
| `/home/how-to-use-these-docs/` | `xref:home:how-to-use-these-docs:index.adoc` | `xref:home::how-to-use-these-docs.adoc` |
| `/self-managed/get-started/intro/` | `xref:self-managed:get-started:intro.adoc` | `xref:self-managed::get-started/intro.adoc` |

`/home/how-to-use-these-docs/` is a real published page (`home/modules/ROOT/pages/how-to-use-these-docs.adoc`), so this is reachable, not theoretical. The `maps the docs-site umbrella components` test asserts the broken forms, which means the bug ships with a green test blessing it.

Two ways out: add a ROOT-only set that emits `xref:<component>::<path>.adoc` for these three, or drop them from the map entirely. Dropping them isn't a regression — the pre-PR behavior was equally broken — but either way the test expectations need to change. `labs`, `connect`, `cloud-data-platform`, `agentic-data-plane`, and `streaming` are all genuinely multi-module (verified: `labs` has `docker-compose`, `clients`, `kubernetes`, ...; adp-docs has `get-started`, `cli`, `connect`, ...), so those entries are correct as written.

## 2. Self-qualifying `streaming` gives up version awareness

The converter's output lands *in* the `streaming` component, which is versioned (26.2 on main, plus `v/*` branches). A component-qualified resource ID with no version resolves to the component's **latest** version, so `xref:streaming:manage:kubernetes/manage-resources.adoc` emitted into a `v/*` branch during a CRD or Helm regen links across versions to latest.

Emitting a same-component xref for `streaming` URLs (`xref:manage:kubernetes/manage-resources.adoc`) keeps the current page's version and is strictly safer. You'd still want `streaming` in the map so the version segment gets stripped — just skip the component prefix when the mapped component is the one being generated into.

## 3. Fragment preservation is an undeclared behavior change (minor, scope)

The `url.hash` block changes output for every previously-anchored URL — they were silently dropped before. It's a good fix, but it isn't in the title or the "Fix" section, and it's the kind of change worth naming: anchors on the live site are rendered heading IDs, and they don't always survive as valid AsciiDoc anchors. `262abc6` in this repo was a fix for exactly that mismatch. An unresolvable anchor degrades to the top of the page rather than failing the build, so the risk is low — it just shouldn't arrive unannounced.

## Nits

- `VERSION_SEGMENT_RE` accepts `current|beta|v?\d+\.\d+` but misses three-part versions (`v25.1.2`) and `latest`. Fine for the URL shapes in use; worth a comment saying so.
- The map is a hardcoded snapshot of five repos' `antora.yml`. A comment pointing at the source of truth would help whoever hits the next component rename.
- The test suite is genuinely thorough (26 cases) and the legacy-prefix precedence test is a nice touch — the `hadLegacyPrefix` guard is the subtle part and it's both correct and well documented.

## Version bump needed

This changes `cli-utils/convert-doc-links.js` with no `package.json` bump, and npm is at 5.3.5. Per the repo's publish rule the fix won't reach consumers until a later PR bumps. #244 claims 5.3.6 and #245 claims 5.3.7, so this needs a version assigned before merge.

JakeSCahill:
Closing in favor of #247, which implements the redesign suggested here: the conversion now runs inside the Antora runtime as a `contentClassified` extension that reverses Antora's own xref→URL mapping against the live content catalog, so there's no slug map, no module-vs-ROOT-page guessing, and no index.adoc/start_page edge cases — a URL converts only if it matches a page actually published in the build.

Two things from this thread carried over directly: the fragment-preservation behavior and the docs#1830 reproduction case (now asserted both in unit tests and in an end-to-end Antora build). The dynamic-playbook question dissolved entirely — running in the real build means all content sources are already there.

Thanks @micheleRP for the review findings that motivated the pivot; the ROOT-only umbrella components and start_page cases were exactly the class of guesswork the catalog-backed approach eliminates.

PR #223, opened 2026-07-28
fix(connect): emit fragment anchors matching rendered heading IDs in whats-new links

JakeSCahill:
## Bug

The Redpanda Connect whats-new generator (`updateWhatsNew` in `tools/redpanda-connect/rpcn-connector-docs-handler.js`) builds field-level links by appending the raw field name as the xref fragment:

```js
componentList += `* xref:components:${type}/${name}.adoc#${fieldName}[${name}]\n`
```

Connector field docs render each field as a section heading (for example, `=== `credentials.host_public_key``), and the docs site derives heading IDs with `idprefix: ''` and `idseparator: '-'`. Under those settings, dots in nested field names become hyphens and `[]` array markers are dropped. A raw dotted field name therefore produces a fragment that matches no element on the page, the link loads the page but never scrolls to the field, and Antora does not validate fragments, so there is no build warning.

Live example (already committed in rp-connect-docs `modules/get-started/pages/whats-new.adoc`, SFTP release entry): the target heading for `credentials.host_public_key` on https://docs.redpanda.com/redpanda-connect/components/inputs/sftp/ renders as `<h3 id="credentials-host_public_key">`, so a generator-emitted `xref:components:inputs/sftp.adoc#credentials.host_public_key[...]` fragment would not resolve. (That entry was hand-corrected to the hyphenated form, which is what the generator should have produced.)

## Verifying the real ID scheme

The original report claimed headings render with the Asciidoctor default `idprefix: _` (`#_field_name`). That is not what production does, so the fix targets the verified scheme instead:

- `redpanda-data/docs-site` `antora-playbook.yml` leaves `idprefix`/`idseparator` commented out, but the `add-global-attributes` extension loads `modules/ROOT/partials/global-attributes.yml` from the `shared` branch of `redpanda-data/docs`, which sets `idprefix: ''` and `idseparator: '-'` (this repo's `preview/shared/modules/ROOT/partials/attributes.yml` matches).
- Published pages confirm it. On https://docs.redpanda.com/redpanda-connect/components/inputs/kafka_franz/ the field headings render as `<h3 id="checkpoint_limit">`, `<h3 id="batching-byte_size">`, `<h3 id="sasl-aws-credentials-from_ec2_role">`, and `<h3 id="batching-processors">` (from `batching.processors[]`). No `_`-prefixed IDs exist on the page.
- Cross-checked with Asciidoctor.js 2.x (the engine Antora uses): loading `== `<field>`` headings with `idprefix: ''`, `idseparator: '-'` produces exactly those IDs for all tested field shapes.

Consequences for existing links: simple top-level field fragments such as the 4.100.0 `#global_table` and `#snapshot_mode` entries do match rendered IDs (verified live on the `aws_dynamodb_cdc` and `oracledb_cdc` pages), and the 4.102.0 entries (`#use_batch`, `#checkpoint_namespace`, `#headers`) will match once rp-connect-docs #466 content is published, because those are all top-level fields. The breakage is limited to nested/dotted and array-marker field names, which the generator emitted verbatim.

## Fix

Add a `fieldAnchor()` helper that reproduces Asciidoctor's section ID derivation under the site's attributes:

1. Downcase.
2. Replace characters that are invalid in section IDs (for example, `[` and `]`) with spaces.
3. Convert runs of spaces and dots to a single `-` (idseparator).
4. Trim leading and trailing separators (no idprefix, and Asciidoctor chomps a trailing separator).

Both whats-new table builders (`buildFieldsTable` and `buildChangedDefaultsTable`) now pass field names through this helper when building the xref fragment. The visible field name in the table keeps its original dotted form. Output for simple top-level field names is byte-for-byte unchanged.

Existing committed entries in rp-connect-docs are intentionally out of scope.

## Tests

New coverage in `__tests__/tools/rpcn-connector-summaries.test.js`:

- `fieldAnchor`: simple names, dotted nested names, `[]` array markers, deep mixed paths (`sasl[].aws.credentials.from_ec2_role`), consecutive-separator collapsing, and downcasing.
- `buildFieldsTable`: emitted xref lines for simple, nested, and array fields, including that the visible cell keeps the dotted name while the fragment is hyphenated.
- `buildChangedDefaultsTable`: derived anchor for a nested field.

All anchor expectations were validated against Asciidoctor.js 2.x output with the site attributes and against rendered IDs on published pages.

Full suite: `npm test -- --testPathIgnorePatterns="cli-contract.test.js|integration.test.js"` — 32 suites, 674 tests, all passing.

### DIFF
--- a/__tests__/tools/rpcn-connector-summaries.test.js
+++ b/__tests__/tools/rpcn-connector-summaries.test.js
@@ -276,6 +276,18 @@ describe('fieldAnchor - fragment anchors for field heading IDs', () => {
   test('downcases names, matching Asciidoctor ID generation', () => {
     expect(fieldAnchor('TLS.Enabled')).toBe('tls-enabled');
   });
+
+  test('deletes invalid characters outright, matching InvalidSectionIdCharsRx', () => {
+    // Asciidoctor removes these characters rather than turning them into
+    // separators, so `foo/bar` renders as <h3 id="foobar">.
+    expect(fieldAnchor('foo/bar')).toBe('foobar');
+    expect(fieldAnchor('a:b')).toBe('ab');
+  });
+
+  test('falls back to the raw name when normalization empties it', () => {
+    // A pathological name must not emit a malformed `#[...]` xref fragment.
+    expect(fieldAnchor('///')).toBe('///');
+  });
 });
 
 describe('buildFieldsTable - whats-new field links', () => {
--- a/__tests__/tools/rpcn-connector-summaries.test.js
+++ b/__tests__/tools/rpcn-connector-summaries.test.js
@@ -18,7 +18,7 @@
 // Jest); stub it out since these tests never touch GitHub.
 jest.mock('../../cli-utils/octokit-client', () => ({}));
 
-const { capToTwoSentences, augmentConnectorData, buildCleanOssData } = require('../../tools/redpanda-connect/rpcn-connector-docs-handler');
+const { capToTwoSentences, augmentConnectorData, buildCleanOssData, fieldAnchor, buildFieldsTable, buildChangedDefaultsTable } = require('../../tools/redpanda-connect/rpcn-connector-docs-handler');
 const { generateConnectorDiffJson } = require('../../tools/redpanda-connect/report-delta');
 
 describe('capToTwoSentences - xref and filename protection', () => {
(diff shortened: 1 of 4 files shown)


micheleRP:
I verified `fieldAnchor` against `@asciidoctor/core` 2.2.8 rather than taking the claim on faith — loaded a document with `:idprefix:` and `:idseparator: -` and compared the real section IDs to the helper's output across 12 field shapes.

**Every realistic connector field shape matches exactly:**

| Field | Asciidoctor ID | `fieldAnchor` |
|---|---|---|
| `checkpoint_limit` | `checkpoint_limit` | ✅ |
| `batching.byte_size` | `batching-byte_size` | ✅ |
| `credentials.host_public_key` | `credentials-host_public_key` | ✅ |
| `sasl[]` | `sasl` | ✅ |
| `sasl[].aws.credentials.from_ec2_role` | `sasl-aws-credentials-from_ec2_role` | ✅ |
| `batching.processors[]` | `batching-processors` | ✅ |
| `foo[].bar` | `foo-bar` | ✅ |
| `TLS.Enabled` | `tls-enabled` | ✅ |

Good fix — these links have been silently landing at the top of the page, and since Antora doesn't validate fragments there was no signal.

## Suggestions

**1. Invalid characters become a space where Asciidoctor deletes them.**

```js
.replace(/[^ \w\-.]+/g, ' ')
```

Asciidoctor's `InvalidSectionIdCharsRx` removes those characters outright, so the two diverge whenever a field name contains something other than word characters, dots, hyphens, or `[]`:

| Field | Asciidoctor | `fieldAnchor` |
|---|---|---|
| `foo/bar` | `foobar` | `foo-bar` |
| `a:b` | `ab` | `a-b` |

No current connector field hits this, so it's not urgent. `.replace(/[^ \w\-.]+/g, '')` matches Asciidoctor exactly, and I confirmed it stays correct for every `[]` case (`sasl[]` → `sasl`, `sasl[].aws` → `sasl-aws`, `batching.processors[]` → `batching-processors`), since the dots still supply the separators.

**2. Worth a comment about collision behavior.** Asciidoctor de-dupes repeated IDs by appending `-2`. If two fields on one page normalize to the same anchor, the second heading gets `-2` and the generated link points at the first. Unlikely with real field names, but the generator can't detect it, so a note near the helper would save someone a puzzling bug report later.

**3. Edge case:** a field name that reduces to an empty string would emit `...adoc#[name]`, a malformed xref. Only reachable with a pathological name — a guard returning the raw name would close it.

## What works well

- The docstring explains *why* the transformation is what it is and cites the live rendered IDs, so the next person doesn't have to re-derive it.
- Test coverage hits every shape that matters, including the mixed array-and-dot deep path.
- Keeping the visible field name in its original dotted form while only rewriting the fragment is the right call, and the tests assert both halves.

micheleRP (approved):
Approving — verified the anchor algorithm against @asciidoctor/core across 12 field shapes and every realistic field form matches exactly. The suggestions in my earlier comment (invalid-char handling for `/` and `:`, the duplicate-anchor note) are non-blocking.

PR #224, opened 2026-07-28
feat: point UI meta tags at JSON attachments that actually exist

JakeSCahill:
## Problem

The top 404s on docs.redpanda.com (~25k requests/day) are versioned JSON attachments that were never generated:

| Resource | Requests/day |
|---|---|
| `/streaming/current/reference/_attachments/redpanda-properties-v26.1.14.json` | 17,658 |
| `/connect/components/_attachments/connect-4.102.0.json` | 6,389 |
| `/streaming/{25.2,25.1,24.3}/reference/_attachments/redpanda-properties-*.json` | ~1,800 |

Root cause: the docs-ui `head-meta.hbs` partial builds `properties-json-url` / `connect-json-url` meta tags from release-tracking attributes (`latest-redpanda-tag`, `latest-connect-version`). Those follow the newest GitHub release, which drifts ahead of the newest generated JSON in the catalog. When `resolve-resource` fails, the template emits a constructed production fallback URL that is guaranteed to 404 — and the property-tooltip script retries it on every page view because failures are never cached. Property tooltips are currently silently broken on `streaming/current`, and the Bloblang tooltip data fetch 404-chains on every page.

The template already prefers `available-properties-tag` / `available-connect-version` (comments say "set by extension to highest version with JSON in catalog") — but that extension was never implemented. This PR implements it.

## What it does

On `contentClassified`, scans attachments and sets per component version (never overwriting values pinned in `antora.yml`):

- `available-properties-tag`: newest `redpanda-properties-<tag>.json` in the `reference` module of each streaming (`ROOT`) version
- `available-connect-version`: newest `connect-<version>.json` in the `components` module of the connect component
- `page-disable-property-tooltips: true` on streaming versions with no properties JSON at all (24.3/25.1/25.2), so tooltips stop fetching a URL that cannot exist

## Testing

- 8 new unit tests (`npx jest __tests__/extensions/set-available-attachment-versions.test.js`)
- Full extension suite passes (60 tests)

## Rollout

After release, register in the docs-site playbook:
```yaml
antora:
  extensions:
    - require: '@redpanda-data/docs-extensions-and-macros/extensions/set-available-attachment-versions'
```
Companion PRs: docs-ui negative caching (incoming), props v26.1.14 + connect 4.102.0 regeneration (already open).

### DIFF
--- a/__tests__/extensions/set-available-attachment-versions.test.js
+++ b/__tests__/extensions/set-available-attachment-versions.test.js
@@ -103,15 +103,18 @@ describe('set-available-attachment-versions extension', () => {
   })
 
   it('sets available-connect-version to the newest connect JSON, comparing numerically', () => {
+    // Production shape: rp-connect-docs is `name: connect` and versionless
+    // (`version: null`), which Antora represents as version '' on both the
+    // component version and the attachment src.
     const compVer = {
-      version: 'current',
+      version: '',
       asciidoc: { attributes: { 'latest-connect-version': '4.102.0' } },
     }
     run(
-      [{ name: 'redpanda-connect', versions: [compVer] }],
+      [{ name: 'connect', versions: [compVer] }],
       [
-        makeAttachment('redpanda-connect', 'current', 'components', 'connect-4.100.0.json'),
-        makeAttachment('redpanda-connect', 'current', 'components', 'connect-4.99.0.json'),
+        makeAttachment('connect', '', 'components', 'connect-4.100.0.json'),
+        makeAttachment('connect', '', 'components', 'connect-4.99.0.json'),
       ]
     )
     expect(compVer.asciidoc.attributes['available-connect-version']).toBe('4.100.0')
--- a/__tests__/extensions/set-available-attachment-versions.test.js
+++ b/__tests__/extensions/set-available-attachment-versions.test.js
@@ -1,0 +1,135 @@
+const { describe, it, expect, beforeEach } = require('@jest/globals')
+
+const extension = require('../../extensions/set-available-attachment-versions.js')
+
+function makeAttachment (component, version, module, relative) {
+  return { src: { component, version, module, family: 'attachment', relative } }
+}
+
+describe('set-available-attachment-versions extension', () => {
+  let mockLogger
+  let handlers
+  let extensionContext
+
+  beforeEach(() => {
+    mockLogger = {
+      info: jest.fn(),
+      debug: jest.fn(),
+      warn: jest.fn(),
+      error: jest.fn(),
+    }
+    handlers = {}
+    extensionContext = {
+      once: jest.fn((event, handler) => {
+        handlers[event] = handler
+      }),
+      getLogger: jest.fn(() => mockLogger),
+    }
+  })
+
+  function run (components, attachments) {
+    const contentCatalog = {
+      getComponents: jest.fn(() => components),
+      findBy: jest.fn(({ component, version }) =>
+        attachments.filter((a) => a.src.component === component && a.src.version === version)
+      ),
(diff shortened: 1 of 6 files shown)


micheleRP:
I traced the whole chain rather than trusting the premise, and it holds up end to end. The diagnosis is right and the fix is in the correct layer.

**Verified in docs-ui:**

1. `head-meta.hbs:30` already prefers `available-properties-tag` over `latest-redpanda-tag`, and `:57` prefers `available-connect-version` — with nothing setting either. The comments in that file even describe the extension that didn't exist yet.
2. `page-disable-property-tooltips` is the name docs-ui already documents (`19-property-tooltips.js:7`).
3. The disable flag genuinely prevents the request rather than just annotating the page — `init()` returns at `19-property-tooltips.js:451` before any `fetch`.

All 8 tests pass locally. Also confirmed the versionless connect path works: `compVer.version` is `''` for `version: null`, and Antora's `findBy` compares with strict equality, so `'' === ''` matches.

## Nits

**1. `extensions/README.adoc` breaks the list's alphabetical order.** The new entry goes in after `add-global-attributes`; the surrounding list is alphabetical (`add-global-attributes`, `add-pages-to-root`, `process-context-switcher`, `validate-attributes`), so it belongs between `process-context-switcher` and `validate-attributes`.

**2. REFERENCE.adoc component naming.** It says "each streaming (`ROOT`) component version" and "the `connect` component", while the code accepts both spellings of each (`['ROOT', 'streaming']`, `['connect', 'redpanda-connect']`). Current reality is `name: streaming` in redpanda-data/docs and `name: connect` in rp-connect-docs. Worth phrasing as "the streaming component (`streaming`, or `ROOT` on older branches)" so the doc doesn't imply `ROOT` is current.

**3. Test fixture models connect as versioned.** The connect test uses `version: 'current'`, but connect is versionless (`version: null` → `''`). The code path is fine; the fixture just doesn't reflect production, and it's the case that carries ~6.4k requests/day.

## Suggestions

**1. The connect side has no equivalent of the disable flag.** If a connect JSON is ever missing from the catalog entirely, `head-meta` still emits `connect-json-url` pointing at the constructed production URL and the 404s continue. Not a problem today (`connect-4.103.0.json` is present), but the guard is one-sided, and the properties side got the fuller treatment.

**2. One fragility worth knowing about.** The tooltip-disable branch is gated on `attributes['latest-redpanda-tag']`, and docs-site#195 registers this extension *before* `version-fetcher/set-latest-version` in all three playbooks (line 368 vs 398 in `antora-playbook.yml`; both hook `contentClassified`, which Antora dispatches in registration order). It works today only because every `antora.yml` pins a `latest-redpanda-tag` fallback. If a version branch ever drops that fallback and relies purely on the fetcher, this branch silently no-ops and the 404s return. Registering after the fetcher, or dropping the gate in favor of the component-name check you already do, would remove the ordering dependency.

## What works well

- Fixing this by reading the catalog instead of adding another release-tracking attribute is the right instinct: it can't drift, because it's derived from what exists.
- Never overwriting an already-set value keeps `antora.yml` authoritative, and there's a test for it.
- `semver.coerce` with `includePrerelease` handles the 4.99.0 → 4.100.0 case that string comparison gets wrong, and the test names that explicitly.
- Logging every attribute it sets means the next person debugging a tooltip 404 can see what the extension decided.

micheleRP (approved):
Approving — traced the full chain and the mechanism works end to end: the attribute reaches `head-meta.hbs` and `init()` returns before any fetch. Findings in my earlier comment are nits (README ordering, REFERENCE.adoc wording, test fixture realism) plus one non-blocking note about the registration-order dependency.

PR #226, opened 2026-07-28
fix(rpk-docs): restore Linux-only platform detection on Linux CI runners

JakeSCahill:
## Problem

When the rpk-docs generator runs on a Linux CI runner (the docs repo's `update-rpk-docs.yml` on `ubuntu-latest`), Linux-only platform detection silently returns nothing, and every command gets stamped `["linux","darwin"]`. In redpanda-data/docs#1831 this flipped 13 Linux-only pages (`rpk iotune` plus the whole `rpk redpanda` tree) to `:page-platforms: linux,darwin`, and the generated data file `docs-data/rpk-v26.2.1-rc2.json` shipped with `linux_only_commands: []`.

Two defects combined to cause this:

1. **The static scan looked in directories that no longer exist.** `detectLinuxOnlyFromSource` only scanned `<source>/pkg/cli/cmd/` and `<source>/cmd/rpk/` for `//go:build linux` tags. Modern rpk keeps commands in `src/go/rpk/pkg/cli/<command>/` (no `cmd` level), so `fs.existsSync` failed for both roots and the function silently returned an empty set. It also only matched explicit tags, missing filename-implied constraints such as `root_linux.go`.
2. **The reliable path can never run in CI.** Dynamic detection (build rpk in a Linux container and natively, diff the two `--print-tree` outputs) is gated on `os.platform() !== 'linux'`. That is inherent: a Linux runner can cross-compile a darwin rpk but cannot execute it to print its tree. So on `ubuntu-latest` only the (broken) static path ran, and the empty set flowed straight into `addPlatformMarkersFromSource`.

## Fix

New module `tools/rpk-docs/detect-platform-commands.js`: a static, platform-independent analysis of Go build constraints that works on any host, including Linux CI.

* Scans `pkg/cli/` (modern layout) plus the legacy roots `pkg/cli/cmd/` and `cmd/rpk/`.
* Treats both explicit constraints (`//go:build linux`, legacy `// +build linux`, including `!`, `&&`, `||`, parentheses, and `unix`) **and** filename-implied constraints (`_linux.go`, `_linux_arm64.go`) as build gates, evaluated per GOOS.
* Detects two gating patterns:
  * **Whole-package gating** - every file in a command package builds only on Linux (e.g. `pkg/cli/iotune`, `pkg/cli/redpanda/tune`).
  * **Dual registration** - a package ships per-platform variants of the same command that register different subcommand sets (e.g. `redpanda.go` with `//go:build linux` registers `start/stop/check/mode/config/tune`, while `redpanda_darwin.go` registers only the hidden `admin` command; same pattern for `root_linux.go`/`root_darwin.go`). The analyzer diffs the constructors referenced in `AddCommand(...)` calls between the linux-buildable and darwin-buildable views of each package, resolving each constructor to its cobra `Use:` name.
* A file with a Linux gate that has a darwin-buildable counterpart in the same package (like `debug/bundle/bundle.go` + `bundle_all.go` with `//go:build !linux`) is correctly **not** treated as Linux-only.
* Deprecated and `Hidden: true` constructors are ignored because `rpk --print-tree` excludes hidden commands (so the deprecated top-level `rpk start/stop/tune/config/mode` aliases don't leak into the results).
* Deliberately conservative: unresolvable references are only trusted when they follow the `New[Xxx]Command` cobra constructor convention, so helper arguments (e.g. `rp.NewLauncher()`) can't produce false Linux-only markings.

Handler changes (`tools/rpk-docs/rpk-docs-handler.js`):

* Replaces the old scanner with the new module (dynamic detection still runs on macOS hosts and merges with the static result, unchanged).
* **Tripwire:** if the combined detection result is empty while the source tree demonstrably contains Linux-constrained files under the scanned roots, a loud multi-line warning is printed naming the files and the likely scan failure. Detection can no longer fail silently.
* The persisted `linux_only_commands` list is now expanded over the actual command tree (each descendant path listed, sorted), matching what dynamic detection has historically produced, so static-only CI runs and Mac dynamic runs emit identical data files for built-in commands.
* New env escape hatch `RPK_DOCS_SKIP_DYNAMIC_DETECTION=1` forces the static-only path, used to validate exactly what a Linux CI runner would produce.

## Ground truth and validation

Ground truth was generated empirically from the `v26.2.1-rc2` tag: rpk built natively on macOS (darwin tree) and in a `golang:1.26.4` Linux container (linux tree), `--print-tree` output diffed. The built-in Linux-only set is exactly 13 paths:

```
rpk iotune
rpk redpanda check
rpk redpanda config (+ bootstrap, init, print, set)
rpk redpanda mode
rpk redpanda start
rpk redpanda stop
rpk redpanda tune (+ help, list)
```

The full generator was then run twice against `--ref v26.2.1-rc2`:

1. **Baseline (Mac, dynamic + static):** dynamic detection finds the built-in set (plus plugin artifacts, see below).
2. **Simulated Linux CI (`RPK_DOCS_SKIP_DYNAMIC_DETECTION=1`, static only):** the new static detection alone produces the same 13 built-in `linux_only_commands` and the same `:page-platforms:` attributes on all built-in pages.

Diff between the two runs for built-in commands: **empty**.

## Testing

* New fixture-based jest suite `__tests__/tools/rpk-docs/detect-platform-commands.test.js` (synthetic source trees covering the modern `pkg/cli` layout, both legacy layouts, explicit tags, `_linux.go` suffixes, the dual-registration redpanda pattern, deprecated/hidden exclusion, the `!linux` counterpart case, and the empty-tree tripwire), plus an integration-style test frozen to the empirically verified rc2 set (7 detection roots, 13 expanded paths).
* Full suite green: 34 suites, 797 tests.

## Notes

* **docs#1831 impact:** the affected pages in that PR are already protected by pinned platform overrides, so no emergency regen is needed there. The next CI regen with this fix produces correct markers on its own.
* **Plugins are out of scope.** Plugin commands (`rpk ai`, `rpk connect`, ...) are not part of the rpk source tree, so static analysis cannot determine their platform support. Dynamic (Mac) runs historically marked `rpk ai *` Linux-only as an artifact of plugins only being installed in the Linux build container, not as a source-level fact. Plugin platform detection is a known follow-up.
* **PR #221 interaction:** #221 touches `rpk-docs-handler.js` at lines ~477-506 (plugin install failure tracking in `fetchRpkTreeFromLinuxSource`) and `generate-rpk-docs.js`. This PR touches neither region (`generate-rpk-docs.js` is untouched), so there is no textual conflict; the two can merge in either order.

### DIFF
--- a/__tests__/tools/rpk-docs/detect-platform-commands.test.js
+++ b/__tests__/tools/rpk-docs/detect-platform-commands.test.js
@@ -1,0 +1,632 @@
+'use strict'
+
+const fs = require('fs')
+const os = require('os')
+const path = require('path')
+
+const {
+  detectLinuxOnlyFromSource,
+  findLinuxConstrainedFiles,
+  warnIfDetectionLooksBroken,
+  fileBuildsOn,
+  goosFromFileName,
+  evaluateBuildExpr,
+  evaluatePlusBuildLines,
+  parseConstructors,
+  parseAddCommandRefs
+} = require('../../../tools/rpk-docs/detect-platform-commands.js')
+
+const { addPlatformMarkersFromSource } = require('../../../tools/rpk-docs/rpk-docs-handler.js')
+
+/**
+ * Write a synthetic source tree: {relativePath: content} pairs
+ */
+function writeSourceTree(root, files) {
+  for (const [relPath, content] of Object.entries(files)) {
+    const fullPath = path.join(root, relPath)
+    fs.mkdirSync(path.dirname(fullPath), { recursive: true })
+    fs.writeFileSync(fullPath, content)
+  }
+}
+
+describe('detect-platform-commands', () => {
+  let tempDir
+
+  beforeEach(() => {
+    tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'rpk-platform-test-'))
+  })
+
+  afterEach(() => {
+    fs.rmSync(tempDir, { recursive: true, force: true })
+  })
+
+  describe('goosFromFileName', () => {
+    test('detects GOOS suffix', () => {
+      expect(goosFromFileName('redpanda_darwin.go')).toBe('darwin')
+      expect(goosFromFileName('bundle_linux.go')).toBe('linux')
+      expect(goosFromFileName('root_windows.go')).toBe('windows')
+    })
+
+    test('detects GOOS_GOARCH suffix', () => {
+      expect(goosFromFileName('cpu_linux_arm64.go')).toBe('linux')
+    })
+
+    test('GOARCH alone does not constrain the OS', () => {
+      expect(goosFromFileName('cpu_arm64.go')).toBeNull()
+    })
+
+    test('non-GOOS words after underscore are not constraints', () => {
+      expect(goosFromFileName('bundle_all.go')).toBeNull()
+      expect(goosFromFileName('bundle_k8s.go')).toBeNull()
+      expect(goosFromFileName('start.go')).toBeNull()
+    })
+
+    test('embedded GOOS word only counts as trailing suffix', () => {
+      // bundle_k8s_linux.go -> linux; linux_thing.go -> no constraint
+      expect(goosFromFileName('bundle_k8s_linux.go')).toBe('linux')
+      expect(goosFromFileName('linux_thing.go')).toBeNull()
+    })
+  })
+
+  describe('evaluateBuildExpr', () => {
+    test('plain GOOS tag', () => {
+      expect(evaluateBuildExpr('linux', 'linux')).toBe(true)
+      expect(evaluateBuildExpr('linux', 'darwin')).toBe(false)
+      expect(evaluateBuildExpr('darwin', 'darwin')).toBe(true)
+    })
+
+    test('negation', () => {
+      expect(evaluateBuildExpr('!linux', 'darwin')).toBe(true)
+      expect(evaluateBuildExpr('!linux', 'linux')).toBe(false)
+    })
+
+    test('boolean operators and parentheses', () => {
+      expect(evaluateBuildExpr('linux || darwin', 'darwin')).toBe(true)
+      expect(evaluateBuildExpr('linux && amd64', 'linux')).toBe(true)
+      expect(evaluateBuildExpr('linux && amd64', 'darwin')).toBe(false)
+      expect(evaluateBuildExpr('(linux || darwin) && !windows', 'linux')).toBe(true)
+    })
+
+    test('unix matches both linux and darwin', () => {
+      expect(evaluateBuildExpr('unix', 'linux')).toBe(true)
+      expect(evaluateBuildExpr('unix', 'darwin')).toBe(true)
+    })
+
+    test('unknown custom tags are unset by default', () => {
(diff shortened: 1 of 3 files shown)


JakeSCahill:
Verified end to end against real rpk v26.2.1 source (sparse clone of redpanda at the tag):

- The old scan roots are demonstrably dead on the real layout: `pkg/cli/cmd` does not exist and `cmd/rpk` holds only main.go, while the commands live in `pkg/cli/<command>` — the exact silent-empty failure described.
- The new detector returns `rpk iotune` plus the six `rpk redpanda` subtree roots, whose descendant expansion is exactly the 13 pages published with `:page-platforms: linux` today. Ran it on macOS and inside a `node:20-slim` Linux container: identical results, confirming platform independence (the CI case).
- Full suite on the branch: 797/797 green. Also merge-tested with #225's current head: one trivial import conflict in rpk-docs-handler.js, combined suite green.

One finding that raises this PR's priority: the v26.2.1 snapshot our validation run produced this week carries `platforms: [linux, darwin]` on `rpk iotune` in the tree and an empty `linux_only_commands` list — the darwin comparison build failed during that run, so dynamic detection never executed and the static path returned nothing, exactly this bug. The published pages are only correct because hand overrides in `rpk-overrides.json` carry `platforms: [linux]`. Any newly added Linux-only command would ship mislabeled until this merges, so it belongs in the train before the next full regeneration.

micheleRP (approved):
Reviewed and tested. Approving, with one sequencing item.

I validated the ground-truth claim independently rather than taking the PR description's word for it: sparse-cloned rpk at `v26.2.1-rc2`, ran `detectLinuxOnlyFromSource` against `src/go/rpk`, and fed the result through `addPlatformMarkersFromSource` with the committed `rpk-v26.2.1-rc2.json` tree. Result: 7 detected roots expanding to exactly the 13 paths you list, zero missing and zero extra.

```
detected roots: 7
expanded linux_only_commands: 13
  rpk iotune
  rpk redpanda check
  rpk redpanda config (+ bootstrap, init, print, set)
  rpk redpanda mode
  rpk redpanda start
  rpk redpanda stop
  rpk redpanda tune (+ help, list)

missing (0): []   extra (0): []
```

That is the check that matters, because `detectLinuxOnlyFromSource` is platform-independent, so what I ran on a Mac is what a Linux runner produces. To be explicit about the limit: I did not exercise a real Linux runner, I exercised the static path that CI actually uses, which is the path that was returning nothing. Tests: 797 green. The union of detected roots and marked tree paths in `linux_only_commands` is a good touch, and so is the tripwire, since silent empty detection was the root cause the first time.

## Sequencing: this conflicts with #225

I test-merged both open generator PRs onto `feat/rpk-plugin-docs`. #229 merges clean; this one conflicts:

```
CONFLICT (content): Merge conflict in tools/rpk-docs/rpk-docs-handler.js
```

It is one hunk, just the `require` lines at the top of the file. Resolution is the union of both import lists plus the new `detect-platform-commands` require. Worth rebasing after #225 merges so the resolution is yours rather than a merge-queue guess.

Also, no version bump here, so after #225 publishes 5.3.0 this needs a bump for the change to reach consumers.

JakeSCahill:
Rebased onto main post-#225 with the one-hunk import union (exactly the resolution from the merge-order note). Full suite 895/895 green. Ready to merge.

PR #229, opened 2026-07-28
feat(rpk-docs): gate subcommand table rows by cloud availability

JakeSCahill:
## Problem

The rpk generator emits subcommand tables with no `env-cloud` gating, while the flags table already supports `cloudOnly`/`selfHostedOnly` conditionals. Cloud-docs single-sources these tables, so rows for commands whose pages exist only in the self-managed docs produced 20+ `target of xref not found` errors in every full-site build. Separately, the `rpk cloud` and `rpk security secret` rows in regular pages emitted in-component xrefs even though their pages are routed to the cloud partials directory and published only by cloud-docs, so those links never resolved.

## Changes

- `templates/command.hbs`: subcommand rows honor `cloudOnly`/`selfHostedOnly` and get wrapped in `ifdef::env-cloud[]`/`ifndef::env-cloud[]`, mirroring the existing flag-row pattern.
- `generate-rpk-docs.js`:
  - The subcommands array carries `cloudOnly`/`selfHostedOnly` from the child command's overrides.
  - New `isCloudSecretCommand()` helper replaces the scattered `startsWith('rpk cloud')`/`startsWith('rpk security secret')` checks (nav skip, cloudSecretDir routing) with strict word-boundary matching.
  - When `cloudSecretDir` is configured, rows for cloud-routed commands in regular pages link across to the `cloud-data-platform` component. Tables inside those families keep in-component xrefs since they render in cloud-docs itself.
- Tests: four new integration tests covering row gating (ifdef/ifndef), the cross-component xref, in-family tables staying in-component, and the no-`cloudSecretDir` case. Full rpk-docs suite passes (293 tests).

## Verification

Ran the updated generator against the real `rpk-v26.2.1-rc2.json` tree with `selfHostedOnly` overrides for the 22 cloud-unavailable commands (redpanda-data/docs branch `rpk-cloud-availability-overrides`): the regenerated parent pages are byte-identical to the hand-fixed pages that shipped in redpanda-data/docs#1837.

The overrides schema already allows both keys at command level, so no schema change is needed.

### DIFF
--- a/__tests__/tools/rpk-docs/override-features.test.js
+++ b/__tests__/tools/rpk-docs/override-features.test.js
@@ -924,4 +924,153 @@ Original fields content that should be replaced.`,
       expect(fs.existsSync(path.join(secretDir, 'some-partial.adoc'))).toBe(true)
     }, 30000)
   })
+
+  describe('subcommand table env-cloud gating', () => {
+    let outputDir
+    let secretDir
+
+    beforeEach(() => {
+      outputDir = fs.mkdtempSync(path.join(os.tmpdir(), 'rpk-out-'))
+      secretDir = fs.mkdtempSync(path.join(os.tmpdir(), 'rpk-partials-'))
+    })
+
+    afterEach(() => {
+      fs.rmSync(outputDir, { recursive: true, force: true })
+      fs.rmSync(secretDir, { recursive: true, force: true })
+    })
+
+    const clusterTree = () => ({
+      name: 'rpk',
+      description: 'Root command',
+      commands: [
+        {
+          name: 'cluster',
+          description: 'Cluster command.',
+          usage: 'rpk cluster [flags]',
+          flags: [],
+          commands: [
+            { name: 'health', description: 'Health command.', usage: 'rpk cluster health [flags]', flags: [] },
+            { name: 'info', description: 'Info command.', usage: 'rpk cluster info [flags]', flags: [] }
+          ]
+        }
+      ],
+      global_flags: []
+    })
+
+    // Cloud-docs single-sources these tables, so a row whose page only exists
+    // in the self-managed docs must not render in the cloud build.
+    test('wraps selfHostedOnly subcommand rows in ifndef::env-cloud', async () => {
+      await generateRpkDocs({
+        tree: clusterTree(),
+        overrides: { commands: { 'rpk cluster health': { selfHostedOnly: true } } },
+        outputDir,
+        rpkVersion: 'test',
+        pluginVersions: {}
+      })
+
+      const page = fs.readFileSync(path.join(outputDir, 'rpk-cluster', 'rpk-cluster.adoc'), 'utf8')
+      expect(page).toMatch(
+        /ifndef::env-cloud\[\]\n\|xref:reference:rpk\/rpk-cluster\/rpk-cluster-health\.adoc\[`rpk cluster health`\]\n\|Health command\.\nendif::\[\]/
+      )
+      // The ungated sibling stays a plain row.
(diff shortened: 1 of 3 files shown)


JakeSCahill:
Integration-tested this against the current #225 head (610e2d90) since both must land in the same train:

**Merge state:** one textual conflict (the import block in rpk-docs-handler.js, trivial union) plus one semantic collision the tests catch: #225 currently drops `rpk cloud` / `rpk security secret` rows from parent subcommand tables entirely (a cruder fix for the same broken-xref symptom this PR solves properly). Resolution: delete #225's early `startsWith` filter in the subcommands builder and let this PR's `isCloudSecretCommand` + cross-component xref logic own the decision. With that, the combined suite is 899/899 green.

**Real-data verification on the combined branch** (full regeneration from `rpk-v26.2.1-rc2.json` on a docs main checkout): `rpk.adoc` and `rpk-security.adoc` regain their `rpk cloud` / `rpk security secret` rows as `xref:cloud-data-platform:...` links, and the regenerated `rpk-security.adoc` subcommands table is byte-identical to the hand-fixed page that shipped in redpanda-data/docs#1837 — the parity claim holds on top of #225, not just against main. 343 regenerated files: zero unbalanced conditionals/tables/tags, and all 343 parse and convert clean under Asciidoctor.

Note for rollout: the env-cloud row gating produces no output until the `selfHostedOnly`/`cloudOnly` overrides land in the docs repo (the `rpk-cloud-availability-overrides` branch mentioned in the description), so merging this before that data is safe.

micheleRP (approved):
Reviewed and tested. Approving, with one small robustness suggestion.

Gating renders correctly, with the directives on their own lines so the preprocessor handles them before table parsing:

```
ifndef::env-cloud[]
|xref:reference:rpk/rpk-cluster/rpk-cluster-health.adoc[`rpk cluster health`]
endif::[]
```

I rendered the generated page with Asciidoctor under both attribute states: zero messages either way, the table parses, and the rows swap as expected (`health` present and `info` absent without `env-cloud`, the reverse with it). The cross-component xrefs land in exactly the two places they should, `rpk.adoc` for `rpk cloud` and `rpk-security.adoc` for `rpk security secret`, and nothing inside the cloud family picks up the prefix. The three `cloud-data-platform:` references I found under `modules/reference/partials/` are pre-existing hand-written links in override content, unchanged by the run, so your assertion about the cloud family holds. Tests: 768 green.

Worth noting as a design confirmation: cloud-docs' nav already lists `rpk-cluster-info.adoc` and already omits `rpk-cluster-health.adoc`, so row gating brings the subcommand tables in line with nav curation that exists there today, rather than introducing a new axis of divergence.

## Suggestion: reject contradictory flags

Setting both `cloudOnly` and `selfHostedOnly` on the same command emits an unsatisfiable pair, so the row silently vanishes from both builds with no warning:

```
ifdef::env-cloud[]
ifndef::env-cloud[]
|xref:reference:rpk/rpk-cluster/rpk-cluster-health.adoc[`rpk cluster health`]
endif::[]
endif::[]
```

It takes an authoring mistake to trigger, so not a blocker, but silence is the unhelpful part. `validate-overrides.js` looks like the natural home for a mutual-exclusion check, alongside the unknown-command-path errors it already reports.

This merges clean onto #225. No version bump here, so it needs one after #225 publishes 5.3.0. And since it touches `templates/command.hbs`, the next full regeneration after #225 carries a second round of template churn, so landing the two close together keeps that to one reviewable wave.

JakeSCahill:
Rebased onto main post-#225 and applied the integration fix from the earlier comment: the row-dropping filter is removed so this branch's cross-component routing owns the `rpk cloud` / `rpk security secret` rows. Combined suite 866/866 green. Ready to merge.

PR #230, opened 2026-07-28
fix(connect): emit context-switcher as array and pass URL entries through

JakeSCahill:
## Problem

Since #216, every multi-type connector page (`aws_dynamodb`, `aws_s3`, `gcp_cloud_storage`, `memory`, `mongodb`, …) logs `Invalid context-switcher format ... expected array` on every site build, and the Type context-switcher dropdown silently never gets processed on exactly the pages that need it.

Two mismatches between producer and consumer:

1. `generate-rp-connect-info.js` `buildContextSwitcher()` returned a **type-keyed object** (`{"cache": {...}, "output": {...}}`), but `process-context-switcher.js` requires an **array** and bails out otherwise.
2. The entries carry root-relative **pub URLs**, which the processor would then try to resolve as Antora **resource IDs** — fixing only the array shape would have traded the format warning for a `Target page not found` warning per entry.

## Changes

- `buildContextSwitcher()` emits an array of `{name, to}` items (current page's type first, deduped by type) — the shape both the processor and the UI's `context-switcher.hbs` expect.
- `process-context-switcher.js` passes root-relative URL entries through untouched: the UI's `resolve-resource` helper already supports them (its comment names this exact producer), and `generate-rp-connect-info` decorates every variant directly so no target injection is needed.
- New test: URL entries produce no warnings and are left unmodified. Full suite passes.

### DIFF
--- a/__tests__/extensions/process-context-switcher.test.js
+++ b/__tests__/extensions/process-context-switcher.test.js
@@ -133,7 +133,7 @@ describe('process-context-switcher extension', () => {
     // generate-rp-connect-info sets pub.url values (root-relative) for connector
     // Type switchers; these need no resource-ID resolution or target injection.
     mockPage.asciidoc.attributes['page-context-switcher'] =
-      '[{"name": "Cache", "to": "/redpanda-connect/components/caches/aws_dynamodb/"},{"name": "Output", "to": "/redpanda-connect/components/outputs/aws_dynamodb/"}]';
+      '[{"name": "Cache", "to": "/connect/components/caches/aws_dynamodb/"},{"name": "Output", "to": "/connect/components/outputs/aws_dynamodb/"}]';
 
     extension.register.call(extensionContext, { config: {} });
     const handler = extensionContext.on.mock.calls[0][1];
@@ -144,8 +144,8 @@ describe('process-context-switcher extension', () => {
     // Attribute is untouched: no 'current' replacement, no rewritten targets
     const attr = JSON.parse(mockPage.asciidoc.attributes['page-context-switcher']);
     expect(attr).toEqual([
-      { name: 'Cache', to: '/redpanda-connect/components/caches/aws_dynamodb/' },
-      { name: 'Output', to: '/redpanda-connect/components/outputs/aws_dynamodb/' }
+      { name: 'Cache', to: '/connect/components/caches/aws_dynamodb/' },
+      { name: 'Output', to: '/connect/components/outputs/aws_dynamodb/' }
     ]);
   });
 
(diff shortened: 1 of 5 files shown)


micheleRP:
Traced the producer/consumer contract through to docs-ui to confirm the pass-through is actually safe, and it is. Good fix — the format warning and the silently-dead dropdown had the same root cause.

**Verification:** docs-ui `src/helpers/resolve-resource.js:62-63` returns `/`-prefixed values unchanged:

```js
if (resource.startsWith('/')) {
  return resource
}
```

That matters for more than just avoiding a warning: `context-switcher.hbs` picks the current item with `{{#if (eq @root.page.url (resolve-resource this.to))}}`. Because `to` is literally the target page's `pub.url` (`generate-rp-connect-info.js:229`), the comparison matches and the toggle renders its label. Skipping resolution is correct rather than merely quiet.

## Suggestions

**1. The test fixture doesn't match production shape.**

```js
'[{"name": "Cache", "to": "/redpanda-connect/components/caches/aws_dynamodb/"}, ...]'
```

`to` comes from `connectPage.pub.url`, and rp-connect-docs is `name: connect`, so production emits `/connect/components/caches/aws_dynamodb/`. `/redpanda-connect/...` is the legacy redirect slug, which `pub.url` never produces. Harmless for the `startsWith('/')` branch, but the fixture is the thing the next reader will treat as documentation.

**2. Undocumented semantics change in the dedup.** The old code let later rows overwrite earlier ones:

```js
data[type.toLowerCase()] = { name: capitalize(type), to: link }   // last wins
```

The new code keeps the first:

```js
if (items.some((item) => item.name === capitalize(type))) continue  // first wins
```

Since `orderedRows` puts `currentRow` first, first-wins is very likely what you want — the current page's own entry survives. Just worth a line in the body or the comment, because it changes behavior for any connector with duplicate types and isn't mentioned.

**3. Commit 2** (`ci: retrigger Netlify preview with raised Node heap`) is empty — worth squashing out before merge.

## What works well

- Fixing both sides of the contract in one PR, rather than only reshaping the producer and trading one warning for `Target page not found`, is the right scope. The body calls that out explicitly.
- The comment at the `continue` explains why no resolution is needed instead of just that it's skipped.
- Asserting `logger.warn` and `logger.error` were never called, and that the attribute is byte-identical afterwards, tests the actual regression rather than a proxy for it.

micheleRP (approved):
Approving — confirmed the contract end to end: docs-ui `resolve-resource.js` passes root-relative URLs through, so the current-label check works because `to` is literally `pub.url`. Non-blocking: the test fixture uses the legacy `/redpanda-connect/` slug rather than the `/connect/` that `pub.url` emits, the dedup change from last-wins to first-wins is worth a line in the body, and commit 2 is empty.

PR #235, opened 2026-07-31
Add end-to-end testing guidelines for automations

JakeSCahill:
## What

Adds `TESTING_AUTOMATIONS.adoc`, a guide to verifying doc generators and their workflows end to end before merging, and links it from the development guides and testing standards in `CONTRIBUTING.adoc`.

## Why

Validating the rpk plugin docs train (#225) showed that 800+ passing unit tests and several expectation-based test runs (exit codes, file counts, extraction counts) can all be green while the generated docs carry real defects. A ground-truth review found seven generator bugs, including truncated flag descriptions, 208 mislabeled version stamps, a dropped nav entry, and code samples mangled into prose. This guide captures the methodology that found them so the next automation change gets the same treatment without rediscovering the pitfalls.

## What it covers

- The core principle: compare output against ground truth captured from the same source artifact, using an independent checker rather than the generator's own parsers.
- The nine-stage loop: real generation on a pristine consumer-repo clone, ground-truth corpus, mechanical checks, Asciidoctor parse, full Antora build, claim verification against snapshots, consumer-repo diff, idempotence, and a judgment pass over meaning.
- Pitfalls with worked examples from #225: platform-dependent ground truth (darwin help stubs), baseline data gaps masquerading as changes, and collateral damage between fix rounds caught only by corpus diffing.
- Triage discipline: regression vs pre-existing content defect vs checker artifact.
- Workflow testing: bot-token PR creation so checks run, prerelease-safe version sorting, and act safety (act picks up the local gh token and will create real PRs from happy-path runs).

Companion to #225. Mergeable independently.

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -35,6 +35,13 @@ link:mcp/DEVELOPMENT.adoc[**MCP server development**]:: How to work on the MCP s
 * Testing guidelines
 * Security and best practices
 
+link:TESTING_AUTOMATIONS.adoc[**Testing automations end to end**]:: How to verify doc generators and their workflows before merging
++
+* Ground-truth corpora and independent checkers
+* Asciidoctor and Antora validation stages
+* Verifying version claims and release-notes diffs
+* Workflow testing, including act safety
+
 == Quickstart
 
 === Prerequisites
@@ -220,6 +227,7 @@ Open a pull request on GitHub with:
 * Write tests for all new functionality
 * Maintain test coverage
 * Test edge cases and error conditions
+* For doc generators and workflows, run an end-to-end verification against the consumer repos before merging. See link:TESTING_AUTOMATIONS.adoc[Testing automations end to end].
 
 == Documentation standards
 
--- a/TESTING_AUTOMATIONS.adoc
+++ b/TESTING_AUTOMATIONS.adoc
@@ -1,0 +1,157 @@
+= Testing automations end to end
+:toc:
+:toc-title: Contents
+
+Unit tests prove that functions behave. They do not prove that an automation produces correct, well-formed documentation in the consumer repos, or that the workflows driving it actually run. Both failure modes have shipped past green test suites in this repo. This guide describes how to test the whole chain before you merge.
+
+The lessons here come from validating the rpk plugin docs automation (link:https://github.com/redpanda-data/docs-extensions-and-macros/pull/225[#225]), where a full ground-truth review found seven generator defects that 800+ passing unit tests and several expectation-based test runs had missed.
+
+== The core principle: test content against ground truth, not against expectations
+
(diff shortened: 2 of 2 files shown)


micheleRP:
This is a genuinely useful guide — specific, evidence-based, and it names real defects with real numbers instead of offering generic testing advice. The "Failure modes of the review itself" section is the most valuable part, because it's the one nobody writes down.

I spot-checked one of the concrete claims rather than assuming, and it holds:

```console
$ printf '26.3.1\n26.3.1-beta.1\n26.3.2\n26.3.1-rc1\n' | sort -V
26.3.1
26.3.1-beta.1
26.3.1-rc1
26.3.2

$ printf '26.3.1\n26.3.1-beta.1\n26.3.2\n26.3.1-rc1\n' | sed 's/-rc/~rc/; s/-beta/~beta/' | sort -V
26.3.1~beta.1
26.3.1~rc1
26.3.1
26.3.2
```

`26.3.1` really does sort below `26.3.1-beta.1`, so a `tail -1` picks the prerelease — and the tilde normalization fixes it. Good that the guide gives the exact remedy and not just the warning.

## Suggestion

**Three em dashes** in the section added by the second commit; team style avoids them:

1. `...correctly classified as pre-existing — then implicitly treated as out of scope for days...`
2. `...because 244 partials looked plausible — the real number was 311.`
3. `...but adp-docs stubs track the page set — three new pages shipped with zero tree changes...`

A colon or a period reads the same in all three.

## Checked and fine

- Sentence-case headings throughout, descriptive rather than "Step N".
- `:toc:` / `:toc-title:` matches the convention already used by `CONTRIBUTING.adoc` and `CLI_REFERENCE.adoc`.
- All source blocks carry a language specifier (`[,bash]`).
- `link:` targets are correct for repo-root docs rendered on GitHub — Antora xref rules don't apply here.
- The `+` continuation attaching a paragraph to the `Pre-existing content defect::` definition-list item is valid AsciiDoc.
- Cross-linked from `CONTRIBUTING.adoc` in both the development-guides list and the testing-standards bullet, so it's discoverable from where someone would actually look.

## What stands out

- "Do the comparison with an *independent parser*" is the load-bearing insight, and the truncation example makes it concrete — a checker built on the generator's own parser inherits the bug and verifies nothing.
- The platform-dependence IMPORTANT block (30 of 32 warnings traced to darwin ground truth) will save the next person a full day of false findings.
- The `act` CAUTION is the kind of thing normally learned by accidentally creating a real PR.
- "Pre-existing is a classification, not a disposition" is a sharp distinction, and citing DOC-2407 against your own review is more persuasive than any abstract guidance would be.

micheleRP (approved):
Approving — genuinely useful guide, and I spot-checked the `sort -V` claim rather than assuming it: `26.3.1` really does sort below `26.3.1-beta.1`, and the tilde normalization fixes it. Only outstanding item is the three em dashes noted in my earlier comment.

PR #236, opened 2026-07-31
feat(rpcn): warn when regenerating a metadata partial drops published sections

JakeSCahill:
## What

Adds a content-loss guard to the RPCN connector docs generator: before overwriting an existing metadata partial, it compares section headings (fence- and literal-block-aware, level-insensitive) and warns with the names of any sections the regenerated content no longer carries. Regeneration remains authoritative — the file is still overwritten — but the loss is now loud in the generation output, which lands in the auto-docs PR body for the reviewer to act on.

## Why

In Connect 4.103.0's auto-docs run (rp-connect-docs#473), the csv input's metadata partial silently lost its "Output CSV column order" section and full YAML example. The section had been hand-migrated from the page into the partial, but it sits outside the upstream description's `== Metadata` block, so regeneration dropped it without a trace. The generator already handles the fully-removed-section case (blanks the partial with a log); this covers the shrinkage case.

## Changes

- `tools/redpanda-connect/metadata-utils.js`: new `sectionHeadings` and `lostMetadataSections` helpers
- `tools/redpanda-connect/generate-rpcn-connector-docs.js`: warn before overwriting when sections are lost
- `__tests__/tools/metadata-partials.test.js`: unit tests for the helpers plus an integration test reproducing the csv case

All 542 tests in `__tests__/tools/` pass.


Feediver1 (commented):
Thanks for adding this guard — the core idea is sound and the heading-level tests are solid. I ran a deep review (each finding below was adversarially verified against the PR head, with runnable repros where noted). Three items I'd ask for changes on, one strong recommendation, and some smaller notes.

## Requested changes

### 1. The warning never reaches the auto-docs PR reviewer it targets

`generate-rpcn-connector-docs.js:448` emits the loss warning via `console.warn` (stderr), but the documented automation (README.adoc, GitHub Actions integration section) redirects **stdout only** (`> output.log`) and builds the PR body solely from the `<!-- PR_SUMMARY_START/END -->` marker range via `sed`. Even with `2>&1`, the warning lands outside the markers. The generator's return value (`{partialsWritten, draftsWritten, partialFiles, draftFiles}`) carries no lost-section data, and neither `rpcn-connector-docs-handler.js` nor `pr-summary-formatter.js` references it — so a dropped section scrolls by in an Actions log and the content loss merges anyway.

**Suggestion:** accumulate lost-section records into the generator's result and render them as a "Content loss" section inside the PR-summary marker block (keep the `console.warn` as a side channel).

### 2. Missing version bump — merging publishes nothing

The PR changes `tools/` code but `package.json` is still `5.2.5`, while npm is already at `5.3.1`. Per the repo's release rules (CLAUDE.md, "Releasing a new version"), tools changes need a version bump in the same PR, and publish only fires when the version exceeds npm. As-is, this guard never ships to consuming repos. Needs at least `5.4.0` plus `npm install --package-lock-only`, committing both files.

### 3. The worst loss case bypasses the warning entirely

`lostMetadataSections` only runs in the `if (metadataOut.trim())` branch. When upstream drops the **whole** `== Metadata` section, the `else if (fs.existsSync(mPath))` branch (`generate-rpcn-connector-docs.js:461`) blanks the partial with `STALE_METADATA_PARTIAL` and emits only a routine-toned `console.log` that names nothing — so total loss of hand-migrated sections gets a *weaker* signal than partial loss. `lostMetadataSections(oldContent, '')` correctly returns all old headings, so wiring it in is a two-line change. The existing tests cover this branch but only assert file blanking, not warning content.

## Strong recommendation

### 4. Fence check runs before `----` block state in `sectionHeadings` (repro included)

`metadata-utils.js:133` — a line starting with <code>```</code> *inside* a `----` listing block sets fence state; the closing `----` is then swallowed by `if (fence) continue`; fence never clears, and every subsequent heading becomes invisible. Repro against the shipped function:

```js
sectionHeadings('== Metadata\n----\n```\n----\n=== Real section\ntext')
// → ["Metadata"]  ("Real section" is lost)
```

Both failure directions reproduce via `lostMetadataSections`: real losses silently missed, and phantom losses reported. Rare input shape, but the consequence is silent and unbounded — exactly the failure class this PR guards against.

Related: this is now the **third** hand-rolled verbatim-region scanner in `tools/redpanda-connect`, each with different rules — `locateMetadata` (not fence-aware), `normalize-metadata.js` (indented fences OK, exact-marker close), and the new `FENCE_DELIMITER` (column-0 only, 3+ chars, any same-char close). Verified disagreement: an indented fence that `normalizeMetadataBlock` writes verbatim is invisible to `sectionHeadings`, producing a spurious loss warning. The `~~~` alternative is untested for `sectionHeadings` and is the only reason fence state is a char rather than a boolean. One shared block/fence-aware line walker would fix both this and #4.

## Minor notes

- **Heading-granularity limitation is undocumented:** prose or fenced examples directly under `== Metadata` with no subheading, or a subsection whose body is gutted while the title survives, produce identical heading sets → no warning (both reproduced). The warning text doesn't overpromise, but a one-line comment acknowledging the scope would help.
- **CRLF input silently disables the guard:** with `\r` on lines, both `BLOCK_DELIMITER` and the heading regex fail, so a CRLF old partial yields zero headings and no warning can ever fire. Low realism (all writers emit LF), but `normalize-metadata.js` trims lines before matching and this file doesn't — trimming would close it.
- **`====` / `....` / `|===` regions aren't treated as verbatim**, so heading-shaped lines inside them count as sections (reproduced; needs unusual content to matter).
- **Test scaffolding:** the new describe block is the third ~25-line copy of the tmpDir/chdir/beforeAll-afterAll boilerplate in `metadata-partials.test.js` — a small shared `setupGeneratorRun()` helper would pay for itself.
- **Follow-up issue (out of scope here):** `migrate-metadata-to-partials.js:84-114`'s "exists, kept" branch discards the page's inline block with no comparison or warning — the largest loss window (first migration) has no guard, and the generator-side guard structurally can't catch it.

Also checked and *not* issues: the `existsSync`/`readFileSync` sequence (any realistically unreadable partial was already unwritable pre-PR), and rename false-positives (the baseline is overwritten each run, so a rename warns exactly once — arguably useful).

micheleRP:
Good guard to add — the csv input case is the kind of loss that's invisible until a reader notices the example is gone. Two things before merge: one real correctness bug (CodeRabbit found the same one), and a question about whether the warning actually reaches a human.

## Fix before merge

**Fence detection runs before the literal-block state, so a fence-like line inside `----` swallows later headings.**

```js
const fenceMatch = line.match(FENCE_DELIMITER);
if (fenceMatch) { ... continue; }
if (fence) continue;
if (BLOCK_DELIMITER.test(line)) { inBlock = !inBlock; continue; }
```

A ``` line inside a `----` literal block sets `fence`, and every heading after that block is skipped until a matching fence appears. `lostMetadataSections` then sees fewer old headings than really exist and reports no loss — so the generator overwrites published content **silently**, which is exactly the failure this PR exists to catch. Swapping the order so the `inBlock` check comes first fixes it. Agreed with CodeRabbit on this one, including the regression fixture it suggests (fence-like line inside `----`, then a real heading afterwards).

## Suggestion

**The warning probably won't reach the reviewer.** The body says the loss "lands in the auto-docs PR body." rp-connect-docs `.github/workflows/update-docs.yml` does capture both streams (`> full_output.txt 2>&1`, line 208), but the PR body only embeds the tail, inside a collapsed block:

```yaml
echo "<details><summary>📋 Generation Output (last 200 lines)</summary>"
...
if [ "$TOTAL_LINES" -gt 200 ]; then
  tail -200 full_output.txt
```

Metadata partials are written mid-run across ~130 connectors, so a full regeneration will very likely push these warnings out of the last 200 lines — and even when they survive, they're behind a `<details>` toggle.

There's already a better channel: `tools/redpanda-connect/pr-summary-formatter.js` emits the `<!-- PR_SUMMARY_START -->` block, which the workflow extracts (lines 239-241) and places **at the top** of the PR body, uncollapsed. Routing content-loss warnings there would make the guard actually load-bearing.

## Minor

**Title-only comparison means an upstream rename reads as a loss.** Fine for a warn-only guard, but the message asserts content "drops," so a reviewer seeing a renamed section may go hunting for content that was never lost. Worth softening the wording to mention rename as a possibility.

## What works well

- Level-insensitive comparison is the right choice, and the test for it documents why (`==` in a description vs `===` in a page-migrated partial).
- Keeping regeneration authoritative and only making the loss loud is the correct call — a guard that blocked the write would stall the whole auto-docs train on a judgment question.
- The integration test that seeds a real published partial on disk, runs the generator, and asserts both the overwrite *and* the warning is much stronger than unit-testing the helpers alone.
- `BLOCK_DELIMITER` has no `/g` flag, so the repeated `.test()` calls are stateless. Worth noting since that pattern often hides a `lastIndex` bug.

JakeSCahill:
Both findings fixed (8df277f), rebased onto main — combined suite 965 green.

- **Fence-state layering**: delimiter states are now layered — inside a `----` literal block only its closer matters, and inside a fence only the matching fence closer matters. Added tests for both directions, including your exact case (a markdown fence inside a literal block followed by a real heading).
- **Warning visibility**: loss warnings now ride the diff object and render as a GitHub warning callout at the top of the PR summary body, before the update header, in both single-version and multi-version formats. console.warn stays as a belt-and-braces channel. Test asserts the block leads the summary.

micheleRP (approved):
Good, well-scoped tripwire, and the right call to keep regeneration authoritative rather than trying to merge. The layered fence/literal-block state machine is correct, and the test asserting that a fence-like line *inside* a `----` block doesn't swallow later headings is exactly the test worth having — that's the failure mode that would have made this guard silently useless.

Approving. The items below are minor and can be follow-ups.

## Minor

**1. Delimiter trimming disagrees with #227.** `sectionHeadings` tests `BLOCK_DELIMITER.test(line)` on the raw line, so `----` with trailing whitespace leaks block state and every heading after it is missed. #227's copy of the same logic in `renderConnectDescription.js` uses `line.trim()`. Untrimmed matches the existing `locateMetadata` in this file, so it's internally consistent — but the two PRs should agree on one behavior before both land.

**2. Upstream heading renames report as losses.** `lostMetadataSections` is title-based, so a rename upstream produces the full `> [!WARNING] This update deletes previously published metadata sections` banner. Since that banner is positioned to stop a merge, a rename generating it will erode trust in it fairly quickly. Consider softening the wording, or detecting the rename case (old title absent + new title present, same position) and reporting it separately.

**3. Misplaced JSDoc.** The new `renderLostSectionWarnings` doc block is inserted *between* `generateMultiVersionPRSummary`'s existing JSDoc and the function itself, so that function loses its docs and `renderLostSectionWarnings` ends up with two blocks above it. #227 makes the identical mistake at the identical spot, which will read as deliberate once both land.

## Nits

- `summaryDiff.lostSectionWarnings = lostSectionWarnings` mutates the diff object in place. Harmless today, but if that object is ever serialized to the diff artifact the field leaks into it.
- The new `describe('PR summary leads with content-loss warnings')` block is glued to the previous `describe` with no blank line between them.
- `sectionHeadings` doesn't track `////` comment blocks or `....` literal blocks. Both are unlikely in a metadata partial; noting for completeness rather than asking for a change.

## Cross-cutting with #227

These two PRs collide in three files and independently invent the same mechanism:

- **Conflicts:** `metadata-utils.js` (both append to the same `module.exports`), `pr-summary-formatter.js` (both insert a render helper at the same location), `rpcn-connector-docs-handler.js` (both rewrite the same `printPRSummary` block).
- **Duplicated design:** both bolt an ad-hoc array onto the diff object (`lostSectionWarnings` / `descriptionReports`), both mutate `masterDiff || diffJson`, and both carry the same "ride the diff object so it doesn't die in the collapsed log" comment. One shared reports channel with a severity field would stop a third copy appearing next time.
- **Duplicated constant:** `BLOCK_DELIMITER` now exists in three places across this PR, #227, and `metadata-utils.js`. Worth exporting the one in `metadata-utils.js`.

Suggest landing this one first — it's the smaller of the two and independently valuable — and having #227 absorb the conflicts and unify the reporting channel.

## Version bump needed

This changes tool code with no `package.json` bump, and npm is at 5.3.5. Per the repo's publish rule the change won't reach consumers until some later PR bumps. #244 claims 5.3.6, #245 claims 5.3.7, #227 claims 5.4.0 — this needs a version assigned before merge.

PR #237, opened 2026-07-31
Honor --path when locating the Redpanda source in the property extractor

JakeSCahill:
## Problem

`find_redpanda_source()` searches hardcoded cwd-relative locations (`tmp/redpanda`, `../tmp/redpanda`, ...) and ignores `--path`, the argument that tells the extractor exactly where the Redpanda source lives. Run the extractor from any directory that does not match one of the guesses and the ConstantResolver plus the constexpr fallback search silently never initialize: no validator enum extraction, no enterprise enum metadata, no constexpr resolution — with only a debug-level log line as evidence.

Concretely: `sasl_mechanisms` ships without `items.enum` and `items.x-enum-metadata`, and `test_enterprise_properties.py::RestrictedOnlyTest::test_sasl_mechanisms_restricted_values` fails against such a build. This surfaced while end-to-end testing #228/#232/#233/#234 with a shared Redpanda clone outside the tool root: the test failure initially looked like a Redpanda dev refactor (`sasl_mechanisms.h`), but the dev source resolves perfectly — the extractor just never looked at it. The Makefile path only works by accident of `cd $(TOOL_ROOT)` plus the clone landing at the hardcoded location.

## Fix

- `main()` records `--path` via a new `set_redpanda_source()`, and `find_redpanda_source()` prefers it, keeping the old search as a fallback for direct module use.
- When the resolver still cannot initialize, the run logs a loud warning naming the exact degradation (which properties lose what) instead of a debug line — the same no-silent-failure treatment as #226's detection tripwire.
- Regression tests for the override, the missing-path fallback, and the no-override search.

## Verification

Real extraction of a Redpanda dev checkout located outside every hardcoded guess, from the tool root (`tmp/` verifiably contains no clone):
- Before: `sasl_mechanisms.items` = `{"type": "string"}` — no enum, no metadata.
- After: full enum `[GSSAPI, SCRAM, OAUTHBEARER, PLAIN]` with `GSSAPI`/`OAUTHBEARER` marked `is_enterprise: true` (matching `enterprise_sasl_mechanisms` in `src/v/config/sasl_mechanisms.h`, the declared source of truth), 654 properties total.
- `test_enterprise_properties.py`: 23/23 pass against the regenerated fixture. Full extractor pytest suite green.

## Relationship to open property PRs

Unblocks the CI concern raised while testing #228/#233/#234 (any extractor-touching PR would hit the failing enterprise test whenever the source lands outside the hardcoded path). No overlap with #228/#232/#233/#234 hunks — merges cleanly alongside all four.

### DIFF
--- a/__tests__/tools/property-extractor/test_source_path.py
+++ b/__tests__/tools/property-extractor/test_source_path.py
@@ -1,0 +1,45 @@
+"""
+find_redpanda_source must honor the --path argument.
+
+Before this, the ConstantResolver initialization and the constexpr fallback
+search hunted hardcoded cwd-relative locations (tmp/redpanda, ...). Running
+the extractor with the source anywhere else silently disabled validator enum
+extraction and enterprise enum metadata: sasl_mechanisms shipped without
+items.enum or x-enum-metadata and only a debug log hinted why.
+"""
+import os
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+sys.path.insert(
+    0,
+    str(Path(__file__).parent.parent.parent.parent / "tools" / "property-extractor"),
+)
+
+import property_extractor
+
+
+class FindRedpandaSourceTest(unittest.TestCase):
+    def tearDown(self):
+        property_extractor.set_redpanda_source(None)
+
+    def test_explicit_path_wins(self):
+        with tempfile.TemporaryDirectory() as tmp:
+            property_extractor.set_redpanda_source(tmp)
+            self.assertEqual(property_extractor.find_redpanda_source(), tmp)
+
+    def test_missing_explicit_path_falls_back(self):
+        property_extractor.set_redpanda_source("/nonexistent/redpanda")
+        result = property_extractor.find_redpanda_source()
+        self.assertNotEqual(result, "/nonexistent/redpanda")
+
+    def test_no_override_uses_search(self):
+        property_extractor.set_redpanda_source(None)
+        # Whatever the search finds, it must not raise
+        property_extractor.find_redpanda_source()
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@redpanda-data/docs-extensions-and-macros",
-  "version": "5.3.1",
+  "version": "5.3.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@redpanda-data/docs-extensions-and-macros",
-      "version": "5.3.1",
+      "version": "5.3.2",
       "license": "ISC",
       "dependencies": {
         "@asciidoctor/tabs": "^1.0.0-beta.6",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@redpanda-data/docs-extensions-and-macros",
-  "version": "5.3.1",
+  "version": "5.3.2",
   "description": "Antora extensions and macros developed for Redpanda documentation.",
   "keywords": [
     "antora",
--- a/tools/property-extractor/property_extractor.py
+++ b/tools/property-extractor/property_extractor.py
@@ -321,16 +321,35 @@ _constexpr_cache = ConstexprCache()
 _type_definitions_cache = {}
 
 # Import topic property extractor
+# Set from --path in main() so every consumer of the Redpanda source uses the
+# tree the caller pointed at, regardless of the current working directory. The
+# hardcoded search below is only a fallback for direct module use.
+_redpanda_source_override = None
+
+
+def set_redpanda_source(path):
+    """Record the Redpanda source directory passed via --path."""
+    global _redpanda_source_override
+    _redpanda_source_override = path
+
+
 def find_redpanda_source():
     """
-    Locate the Redpanda source directory by searching standard locations.
-    
-    The property extractor looks for the Redpanda source code in multiple
-    locations to handle different execution contexts (project root, tools directory, etc.).
-    
+    Locate the Redpanda source directory.
+
+    Prefers the path passed via --path (recorded by set_redpanda_source);
+    falls back to searching standard locations relative to the current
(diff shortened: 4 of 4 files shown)


Feediver1 (commented):
## Code review (10 findings: 7 confirmed, 2 plausible, 1 trivial)

### Critical

**1. The fix reintroduces the exact bug it fixes, for the README-documented invocation.** (`property_extractor.py:330`) — CONFIRMED
The override returns the raw `--path`, but every consumer appends `src/v` to it (`transform_files_with_properties` :723, constexpr fallback :588–592, `build_cache` :165–170). The README documents `--path tmp/redpanda/src/v` (README.adoc:847, 959, 982), and `main()` itself normalizes exactly these shapes at :2800–2806 — then discards the normalized root. With the documented invocation, consumers compute `tmp/redpanda/src/v/src/v` (nonexistent) → `constant_resolver` stays `None` → `sasl_mechanisms` ships without `items.enum`/`x-enum-metadata` again. Before this PR, the cwd fallback returned `tmp/redpanda` and worked. **Fix:** normalize inside `set_redpanda_source()`/`find_redpanda_source()` (extract the :2800–2806 logic into a shared helper).

**2. The fix doesn't fully work even under `make build`: a second module copy never sees the override.** (`property_extractor.py:2871`) — CONFIRMED
The Makefile runs the script as `__main__`, but `transformers.py:620` later does a deferred `from property_extractor import …`, loading a *second* copy of the module whose `_redpanda_source_override` is `None` and whose `_constexpr_cache` is unbuilt. Constexpr resolution through `FriendlyDefaultTransformer` (transformers.py:1583–1585, 1764/1839) therefore still uses the old cwd guesses and silently emits raw C++ identifiers as defaults. Net effect: enum extraction looks fixed (runs in `__main__`) while constexpr-valued friendly defaults stay broken — a partial, hard-to-notice failure under the sanctioned entry point.

**3. No `package.json` version bump — the fix never ships.** (`package.json:3`) — CONFIRMED
Per repo CLAUDE.md ("Releasing a new version"), tool changes publish to npm only when `version` exceeds what's published. The diff leaves `5.2.5` untouched, so on merge the Action skips publishing and downstream repos keep the broken extractor. Bump to `5.2.6` + `npm install --package-lock-only`, commit both.

### Major

**4. The new loud warning misdiagnoses its likeliest trigger, and the sibling failure stays silent.** (`property_extractor.py:729`) — CONFIRMED
With a `src/v`-shaped `--path`, the warning says "Pass `--path` … to fix this" to a user who already passed it (it's `required=True`), pointing away from the real fix. Meanwhile `build_cache(options.path)` skips every missing search dir silently and reports "✅ Cached 0 constexpr identifiers" as a success-styled INFO line — unresolved defaults ship with exit code 0.

**5. A set-but-missing override is silently discarded.** (`property_extractor.py:333`) — PLAUSIBLE (library-only; dead code from the CLI since `validate_paths()` exits first)
A direct-module caller with a typo'd path silently falls back to `tools/property-extractor/tmp/redpanda` — in CI, a date-keyed cached clone of **dev** (test.yaml:238–244) — extracting from the wrong Redpanda version with zero diagnostics. `test_missing_explicit_path_falls_back` locks this in as intended behavior.

**6. The test suite provides no regression barrier for the fix.** (`test_source_path.py:34`) — CONFIRMED
Deleting `set_redpanda_source(options.path)` from `main()` — reverting the entire fix — leaves all three tests green. One test has no assertion, one is satisfied by `None`, and `test_explicit_path_wins` round-trips an empty temp dir with no `src/v` subtree (exactly the value shape that breaks production per finding 1).

**7. Module-scope `import property_extractor` makes pytest collection require the full import chain.** (`test_source_path.py:21`) — PLAUSIBLE
In a minimal local venv (no `pyyaml`), collection of this file errors out the whole directory's run — the existing tests avoided this by importing lighter modules or deferring imports. `importorskip` or importing in `setUp` contains the blast radius. CI is unaffected.

### Suggestions

**8. Thread the (normalized) path as a parameter instead of a mutable module global.** (`property_extractor.py:307`) — the global is a second source of truth its own consumers bypass (`build_cache` still gets raw `options.path` directly), it makes the module non-reentrant, forces test teardown, and creates an ordering hazard: any source-tree read added above :2871 silently gets the old cwd-guess behavior back.

**9. Performance: no negative-result memoization on the newly-activated resolution paths.** (`property_extractor.py:563`) — where the fallback previously bailed in microseconds (source not found), it now walks five `src/v` subtrees per cache-missed identifier, and `ConstantResolver` caches successes only — repeated whole-tree walks across ~650 properties. Treat a miss in the pre-built cache as authoritative, or memoize misses.

**10. Trivial cleanup while the hunk is open:** the stranded `# Import topic property extractor` comment now sits atop the unrelated override block (:303), and the new test carries an unused `import os`.

### What works well

- The diagnosis and the no-silent-failure intent are exactly right — findings 1, 2, and 4 are about finishing that intent, not disputing it.
- `test_explicit_path_wins` is the right shape of regression test; it just needs a realistic fixture.

micheleRP (approved):
Approving. Reviewed alongside #232 and #234 and tested as a group.

The symptom in the docstring is what makes this worth having: `sasl_mechanisms` shipping without `items.enum` or `x-enum-metadata`, with only a debug log hinting why. A silent loss of enterprise enum metadata whenever the source is not in the expected cwd-relative spot is exactly the class of failure that survives for months.

Threading the `--path` value through an explicit `set_redpanda_source` rather than adding another hardcoded search location is the right fix, and keeping the existing search as a documented fallback for direct module use avoids breaking anything that relied on it.

The tests cover the three cases that matter: explicit path wins, a missing explicit path falls back rather than dying, and the no-override path still does not raise. `tearDown` resetting the override is good hygiene, since the state is module-level.

Two notes on the group rather than this PR:

- #232, #234, and #237 are mutually conflict-free in all six orders. The three-way merge is clean and green: 916 JS tests and 72 Python tests, including this PR's `test_source_path.py`.
- None of the three bumps `package.json`, so the merged trio sits at 5.3.1, which is already published, and `publish-to-npm` no-ops when the version is unchanged. Whichever merges last should carry the bump.

PR #239, opened 2026-07-31
Reject unsupported host platforms in the container build fallback

JakeSCahill:
Suggestion 3 from @micheleRP's #238 review: the host-platform mapping sent `win32` to `linux`, silently reproducing the wrong-binary bug class the cross-compile fix eliminated. Unsupported platforms now throw with a message naming `--rpk-bin` as the escape hatch. macOS and Linux behavior unchanged.

No version bump — intended to merge before #228 so the 5.3.1 publish carries it with the rest of the train.

Follow-up to #238.

### DIFF
--- a/tools/rpk-docs/rpk-docs-handler.js
+++ b/tools/rpk-docs/rpk-docs-handler.js
@@ -1386,6 +1386,16 @@ function acquireRpkBinary(rpkVersion, options = {}) {
     // macOS host that needs it for plugin installs (review finding on the
     // 5.3.0 train). rpk builds with CGO disabled, so cross-compilation
     // from the linux container is safe.
+    // Only macOS and Linux hosts are supported: mapping anything else to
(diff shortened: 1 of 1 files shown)


micheleRP (approved):
Approving. This is exactly suggestion 3 from the #238 review, implemented minimally: unsupported hosts throw instead of being mapped to `linux`, the original `nativeErr` message is preserved so the Go-version cause is not lost, and the error names `--rpk-bin` as the way forward. macOS and Linux paths are untouched.

Two nits, neither blocking:

1. The check sits after `console.warn('...building in a container...')`, so a Windows user is told a container build is starting and then told the platform is unsupported. Moving the guard above that warn would avoid announcing work that never happens.
2. No test. Stubbing `process.platform` is awkward enough that I would not insist, and the code is a three-line guard, so this is optional.

Sequencing: since this has no version bump, it needs to merge before #228 so the 5.3.1 publish carries it.

PR #240, opened 2026-08-03
Warn when a rerender drops a page's Flags section

JakeSCahill:
Implements @micheleRP's suggestion from the merge-order review, which came with direct evidence: with the curated `rpk connect run` Flags override removed (#1865) and a snapshot carrying no extracted connect flags, the page renders title → Usage → Global flags with **no Flags section, exit 0, and zero warnings** — nothing catches the class at generation time, and the resulting diff reads like intended cleanup.

The generator now compares each page against the file it is about to overwrite: a page that previously had a `== Flags` section and loses it in the new render emits a warning naming the two likely causes (a snapshot without flag extraction, or a curated Flags override removed before the extracted table exists). Warning-only by design — there are legitimate transitions — but it turns a silent page regression into a reviewable log line and a line in the workflow output.

Tested with a first-render/re-render cycle reproducing the connect run shape. Full suite green. No version bump: rides the next publish.

Follow-up from the #1862/#1865 merge-order review.

### DIFF
--- a/__tests__/tools/rpk-docs/generate-rpk-docs.test.js
+++ b/__tests__/tools/rpk-docs/generate-rpk-docs.test.js
@@ -1345,3 +1345,35 @@ describe('placeholder brace escaping', () => {
     expect(out).toContain('`table\\{vbar}json`')
   })
 })
+
+describe('vanishing Flags tripwire', () => {
+  const os2 = require('os')
+
+  test('warns when a rerender drops a previously present Flags section', async () => {
+    const tmp = fs.mkdtempSync(path.join(os2.tmpdir(), 'flags-tripwire-'))
+    const tree = {
+      name: 'rpk',
+      commands: [{
+        name: 'widget',
+        description: 'Widget things.',
+        usage: 'rpk widget [flags]',
+        flags: [{ name: 'verbose', type: 'bool', description: 'Louder.' }],
+        commands: []
+      }]
+    }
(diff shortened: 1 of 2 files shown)


micheleRP (approved):
Approving. This is the tripwire suggested from the merge-order review, and I tested it by replaying the exact incident it exists to catch.

**Hazard replay:** on current docs main, I stripped all connect flags from a snapshot copy (simulating the pre-re-dispatch state) and regenerated with this branch's build. The warning fired once per affected page, naming both likely causes:

```
⚠️  rpk connect run: the previously rendered page had a Flags section and this render has none. ...
```

Previously that exact scenario produced exit 0 and a completely clean log.

**One design property worth documenting** (not a defect): the comparison is against the on-disk page, so the warning fires on the **first** offending render only — a second local run is silent because the first already wrote the flagless pages. In CI this is correct behavior, since every workflow run starts from a fresh checkout of committed state, so the warning keeps firing until either the data is fixed or the bad PR merges. Its job is to stop that first PR, and it does.

Also verified: comparing on-disk state means commands that never had flags stay quiet, so there is no noise floor — a clean regeneration of current main produces zero of these warnings.

Merges cleanly after #241 (`generate-rpk-docs.js` auto-merges), and the combined tree passes 933 tests. Same release note as #241: no bump here, so both ride the 5.3.3 cut.

PR #244, opened 2026-08-04
Release 5.3.6: export the set-available-attachment-versions extension

JakeSCahill:
## What

docs-site#195's preview build fails with `ERR_PACKAGE_PATH_NOT_EXPORTED`: #224 shipped `extensions/set-available-attachment-versions.js` but never added it to the package `exports` map, which enumerates every extension subpath explicitly. The extension loads fine inside this repo (tests require it by relative path), so nothing here could catch it — only a real consumer build does.

- Adds the missing exports entry (audited: it was the only gap).
- Adds a regression guard test asserting every `extensions/*.js` is reachable through `exports`, so this class can't recur.
- Carries the 5.3.6 bump so docs-site#195 can pin past it.

## Validation

985 tests pass. Subpath resolution verified with `createRequire` against the local package.

cc @micheleRP — this is the missing piece behind the docs-site#195 Netlify failures; after publish I'll bump #195's lockfile and its checks should go green.

### DIFF
--- a/__tests__/package-exports.test.js
+++ b/__tests__/package-exports.test.js
@@ -1,0 +1,29 @@
+'use strict'
+
+const fs = require('fs')
+const path = require('path')
+
+/**
+ * Every extension file must be reachable through the package exports map.
+ * The map enumerates subpaths explicitly, so a new extension that ships
+ * without an entry loads fine in this repo but throws
+ * ERR_PACKAGE_PATH_NOT_EXPORTED in every consumer — found the hard way when
+ * set-available-attachment-versions broke the docs-site build despite the
+ * file being present in the published package.
+ */
+describe('package exports cover all extensions', () => {
+  const pkg = JSON.parse(fs.readFileSync(path.join(__dirname, '..', 'package.json'), 'utf8'))
+
+  const exportedTargets = new Set(
+    Object.values(pkg.exports || {}).map((v) => (typeof v === 'object' ? v.require : v))
+  )
+
+  const extensionFiles = fs
+    .readdirSync(path.join(__dirname, '..', 'extensions'))
+    .filter((f) => f.endsWith('.js'))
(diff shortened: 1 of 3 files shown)


micheleRP (approved):
Correct, minimal, and the right fix. The regression guard is a good instinct — this class of bug is invisible inside the repo, so a test that asserts the *contract* rather than the behavior is exactly right.

Approving. Suggest landing this first: it unblocks docs-site#195 and has zero interaction with the rest of the open batch.

## One gap (minor, non-blocking)

The guard globs only top-level `extensions/*.js`, but the exports map already carries subdirectory entries:

```
"./extensions/version-fetcher/set-latest-version": "./extensions/version-fetcher/set-latest-version.js"
"./extensions/algolia-indexer/index":              "./extensions/algolia-indexer/index.js"
```

There are three subdirectories today (`algolia-indexer/`, `util/`, `version-fetcher/`). A new extension added under one of those reproduces exactly the bug this test exists to prevent, and the guard stays green. Recursing the walk (or asserting per-subdirectory) closes it.

## Nit

`typeof v === 'object' ? v.require : v` silently ignores `import` / `default` conditions. Fine for the map as it exists today — worth a one-line comment so the next person adding a dual-condition entry knows the test only checks `require`.

## Version sequencing across the open batch

Flagging because several PRs are in flight and npm is still at 5.3.5:

| PR | Bump | Changes tool code? |
|---|---|---|
| **244** | **5.3.6** | yes |
| 245 | 5.3.7 | yes |
| 227 | 5.4.0 | yes |
| 222 | none | yes |
| 236 | none | yes |
| 246 | none | yes |

If this lands first, 5.3.6 publishes cleanly and #245's 5.3.7 follows without a re-bump. 222/236/246 will need bumps assigned before they merge, or their changes sit unpublished until the next PR that does bump.

PR #245, opened 2026-08-04
feat(rpk-stubs): copy the partial description into created stub headers

micheleRP:
## What this does

Stub pages created by `doc-tools generate rpk-plugin-stubs` now carry the partial's `:description:` in their own header.

## Why

The generated rpk partials repeat `:description:` inside their `single-source` tag, intending single-sourcing consumers to inherit the meta description (see the comment in `command.hbs`). That inheritance never works: Antora resolves page metadata with a header-only parse that stops at the stub's first blank line, so the `include::` is never evaluated for attributes. Every stub page created by the reconciler ships the consumer site's generic fallback meta description instead — e.g. all 119 rpk stub pages in adp-docs before redpanda-data/adp-docs#189 backfilled them.

This change fixes it at the source for future stubs: `readPartialTitles` captures the partial's description and `renderStub` writes it into the stub header, immediately after the title. Partials without a description produce stubs without the line. Existing stubs are untouched (the reconciler only creates and deletes).

## Testing

Two new cases in `__tests__/tools/rpk-docs/plugin-stubs.test.js`, written failing-first: the description is copied into the created stub header in position, and no `:description:` line is emitted when the partial has none. Full rpk-docs suite: 448/448 pass.

## Version

Bumps 5.3.5 → **5.3.7** (skipping 5.3.6, which open PR #244 claims). If this merges before #244, that PR will need a re-bump per the repo's publish rules. On merge, the publish action ships to npm and `sync-rpk-ai-stubs.yml` in adp-docs picks it up automatically on its next run (`npx -p @redpanda-data/docs-extensions-and-macros@^5.3.0`), so newly synced stubs will pass adp-docs' new description lint with no further changes.

### DIFF
--- a/__tests__/tools/rpk-docs/plugin-stubs.test.js
+++ b/__tests__/tools/rpk-docs/plugin-stubs.test.js
@@ -70,6 +70,31 @@ describe('plugin stub reconciler', () => {
     expect(stub).toContain('include::streaming:reference:partial$rpk-ai/rpk-ai-auth-login.adoc[tag=single-source]')
   })
 
+  test('copies the partial description into created stub headers', () => {
+    fs.writeFileSync(path.join(partialsDir, 'rpk-ai-agent-create.adoc'),
+      '= rpk ai agent create\n:description: Create an agent.\n:page-platforms: linux,darwin\n\n// tag::single-source[]\n:description: Create an agent.\nBody.\n// end::single-source[]\n')
+
+    const result = run()
+
+    expect(result.created).toEqual(['rpk-ai-agent-create.adoc'])
+    const stub = fs.readFileSync(path.join(stubDir, 'rpk-ai-agent-create.adoc'), 'utf8').split('\n')
+    expect(stub[0]).toBe('= rpk ai agent create')
+    expect(stub[1]).toBe(':description: Create an agent.')
+    expect(stub[2]).toBe(':page-preview: true')
+  })
+
+  test('creates stubs without a description line when the partial has none', () => {
+    fs.writeFileSync(path.join(partialsDir, 'rpk-ai-bare.adoc'),
+      '= rpk ai bare\n\n// tag::single-source[]\nBody.\n// end::single-source[]\n')
+
+    const result = run()
+
+    expect(result.created).toEqual(['rpk-ai-bare.adoc'])
+    const stub = fs.readFileSync(path.join(stubDir, 'rpk-ai-bare.adoc'), 'utf8')
+    expect(stub).not.toContain(':description:')
+    expect(stub).toContain(':page-preview: true')
+  })
+
   test('rebuilds the nav block hierarchically and preserves surrounding nav', () => {
     writePartial('rpk-ai.adoc', 'rpk ai')
     writePartial('rpk-ai-auth.adoc', 'rpk ai auth')
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@redpanda-data/docs-extensions-and-macros",
-  "version": "5.3.6",
+  "version": "5.3.7",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@redpanda-data/docs-extensions-and-macros",
-      "version": "5.3.6",
+      "version": "5.3.7",
       "license": "ISC",
       "dependencies": {
         "@asciidoctor/tabs": "^1.0.0-beta.6",
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "./extensions/generate-rp-connect-info": "./extensions/generate-rp-connect-info.js",
     "./extensions/generate-fields-only-pages": "./extensions/generate-fields-only-pages.js",
     "./extensions/add-global-attributes": "./extensions/add-global-attributes.js",
+    "./extensions/set-available-attachment-versions": "./extensions/set-available-attachment-versions.js",
     "./extensions/git-full-clone": "./extensions/git-full-clone.js",
     "./extensions/add-git-dates": "./extensions/add-git-dates.js",
     "./extensions/add-faq-structured-data": "./extensions/add-faq-structured-data.js",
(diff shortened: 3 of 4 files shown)


JakeSCahill:
Sequencing heads-up: #244 (the missing exports entry that's breaking docs-site#195's build) carries a 5.3.6 bump and is also awaiting review. This PR skips to 5.3.7, which works in either order version-wise, but if this merges first, #244's package.json goes backwards and needs a rebase to 5.3.8 (or a re-point to no-bump riding this release). Cleanest path: merge #244 first (one-line fix plus a regression-guard test), then this rebases trivially — 5.3.6 then 5.3.7, each publish carrying its own change.

JakeSCahill:
@micheleRP following up on the root cause you found: I tested whether stubs can inherit the description **dynamically** instead of by copy — including the single-source region without the blank line. Result: the description arrives, but the body is destroyed (the first content line parses as an author line under the header-only parse). The shape that delivers both is two includes — a nested `tag::meta[]` region around the in-tag description, included in the stub header above the first blank line, with the body include unchanged below.

Prototype with the full experiment table: #246. It composes with this batch rather than replacing it: your backfills fix today's pages and stay correct (a later meta include would override the literal, fresher-wins), your lint holds the line, and #246 closes the freshness gap — cloud-docs' 145 rpk backfills have no sync and rpk descriptions get rewritten by regens (5.3.3 alone rewrote ~36). It does nothing for the 234 allowlisted Connect stubs, which need the editorial effort either way.

Both PRs touch generate-plugin-stubs.js — happy to rebase mine after yours lands, and the version-sequencing note above still applies (#244 → #245 → #246 works cleanly).

JakeSCahill (approved):
Reviewed the extraction and render paths plus both tests, and traced the interactions:

- The first-match `^:description:` extraction is correct against real partials: it takes the header line (including `pass:q[...]` wrappers, which the stub should carry verbatim — adp-docs#189's manual backfills do the same), and single-line values are guaranteed by the generator that produces the partials.
- The no-description branch correctly emits no line rather than an empty attribute, and the failing-first tests pin both shapes.
- Composition with #246 stands as discussed: this PR is the literal-copy fallback half — at my rebase, creation becomes `hasMetaTag ? dynamic include : this PR's literal copy`, and the in-place upgrader strips stale literals when partials gain the meta region (your precedence finding). Nothing here needs to change for that.
- Sequencing note from earlier still applies: #244 (5.3.6) first, then this at 5.3.7 — merging this first leaves #244's version going backwards.

CI green, 5.3.7 bump consistent in package.json and both lockfile fields. Approving.

micheleRP:
Review notes from a pass over the whole open batch (#222, #227, #236, #244, #245, #246), recorded here for the record since I can't formally review my own PR.

## Mechanism confirmed

I re-derived the AsciiDoc reasoning independently rather than taking the premise on trust, and it holds: with a header-only parse the reader stops at the first blank line, so an `include::` below it is never preprocessed and never contributes attributes. The corollary that #246 documents — an include on the line *directly* after the doctitle has its first line consumed as the **author line** — is standard AsciiDoc header parsing, which is why the naive no-blank-line variant destroys the body. Both this PR and #246 are built on correct mechanics.

## Minor

`content.match(/^:description:[ \t]*(.+)$/m)` scans the whole file rather than just the header. In practice the header copy appears first so it wins, but scoping the search to the pre-blank-line header would match the stated intent and stay correct if a future template ever drops the header copy and keeps only the in-tag one.

## Relationship to #246

- **Guaranteed conflict.** Both PRs rewrite the same two sites: the `partials.push({ file, title })` call in `readPartialTitles` and the `renderStub({...})` signature. Whichever lands second rebases.
- Given how small this diff is, folding it into #246 as a single change may be cheaper than rebasing. The composition proposed on #246 (`hasMetaTag ? dynamic include : literal copy`) is the right shape — this PR's literal copy remains the correct fallback for partials generated before 5.4, so it should not be dropped.
- One ordering detail to settle at merge: this PR puts `:description:` immediately after the title and *before* the attribute lines; #246 puts the `[tag=meta]` include *after* them. Both are valid header positions, but the merged version should pick one so created and upgraded stubs look alike.

## Version

5.3.7 skipping 5.3.6 is correct given #244. If #244 lands first this needs no re-bump.

JakeSCahill:
CI is green again — the `test-all` failures were my own doing: yesterday's conflict-resolution rebase took this branch's `package.json` wholesale and silently dropped the `set-available-attachment-versions` exports entry that #244 added on main. The `package-exports` regression guard from #244 caught it, which is a nice validation of that guard. `package.json` on this branch is now byte-identical to main except the 5.3.7 bump.

On the #246 relationship — recommendation after reading #246's current diff: **merge this PR first and rebase #246 on top, rather than folding.**

1. **#246 doesn't currently contain the literal-copy fallback.** Its "older partials without the region" test uses a fixture with no header `:description:` at all, so a pre-5.4 partial that *has* a description but no `tag::meta[]` region would still produce a description-less stub. Folding therefore isn't a squash — #246 would need to absorb this PR's extraction and render logic either way. Doing that as a rebase keeps this PR's approved review and lets #246 stay focused on the dynamic mechanism.
2. **Readiness asymmetry.** This PR is approved and green today and fixes the current gap (stubs created with no description). #246 is larger (11 files) and still in review; folding couples a ready fix to an unfinished one.
3. **The rebase is small.** The two conflict sites are exactly where #246 adds `hasMetaTag` — resolution is "keep both fields, keep both branches." The composed `renderStub` becomes `hasMetaTag ? [tag=meta] include : (description ? literal copy : nothing)` — the shape already agreed here.

On the ordering detail: the conflict is smaller than it looks. Both PRs already put their mechanism directly under the doctitle — this PR's literal `:description:` and #246's `[tag=meta]` include (whose own test asserts line 1, and which dodges the author-line hazard because the include's first line is an attribute entry). So the merged convention can simply be: *the description mechanism — literal or include — is always the first header line after the title, before other attribute lines.* Created and upgraded stubs then look alike, and #246's upgrade path (stripping a stale backfilled literal when adding the include) already prevents the two from coexisting.

I'll rebase #246 with that composition as soon as this merges.

PR #246, opened 2026-08-05
Let single-source stubs inherit descriptions as header attributes

JakeSCahill:
## What

Companion to @micheleRP's description-coverage batch (adp-docs#189, cloud-docs#664, #245): makes stub descriptions **dynamic** so they can never drift when regeneration rewrites them.

## The experiment behind it

Reproduced the root cause in a minimal Antora site built with the production docs-ui bundle:

| Stub shape | Meta description | Page body |
|---|---|---|
| Title, blank line, `include [tag=single-source]` (today) | ❌ generic site text | ✅ |
| Title, include on the next line (no blank) | ✅ inherited | ❌ **destroyed** — first content line parses as an author line |
| Title, `include [tag=meta]`, blank line, `include [tag=single-source]` | ✅ inherited | ✅ |

The header-only metadata parse stops at the first blank line, so the description must arrive as a header attribute — but a bare header include swallows the body. The two-include shape gives both.

## Changes

1. **command.hbs** nests a `tag::meta[]` region around the in-tag `:description:`. Tag directive lines are stripped from tag-filtered includes, so every existing `tag=single-source` consumer renders byte-identically. The old template comment claiming the in-tag copy alone reached stub metadata was wrong (it's where my earlier explanation came from too) — corrected with the verified behavior.
2. **generate-plugin-stubs.js** emits the header include for partials that carry the region, and keeps the plain shape for older partials so builds never warn about a missing tag. Graceful adoption: stubs upgrade as partials regenerate.
3. Tests for the template region, both stub shapes, and the no-warn guard.

## How this composes with the static-backfill batch

- #189/#664 land as-is. ~~Where both a literal and the meta include exist, the include wins.~~ **Corrected (micheleRP's precedence finding, re-verified): the LATER entry wins**, so a literal below the include defeats it. The in-place upgrader therefore strips the backfilled literal from generated reference stubs when it inserts the include — inherently scoped, since the reconciler never touches hand-maintained prose stubs. The cloud-docs conversion (DOC-2414) must do the same, with the same scoping.
- #245's generator change and this one touch the same file — happy to rebase whichever lands second. At rebase, the two compose as `hasMetaTag ? dynamic include : literal copy` (micheleRP's suggestion), so stubs created against pre-5.4 partials keep getting #245's literal description instead of regressing to none.
- The freshness gap this closes: cloud-docs' 145 rpk backfills have no sync mechanism, and rpk descriptions change on regen (the summary fixes in 5.3.3 rewrote ~36 of them).
- No help for the 234 allowlisted Connect stubs — those have no upstream description to inherit; that remains the editorial effort.

No version bump — rides whichever release goes out next (sequencing note on #245 applies).

## Producer coverage (complete)

Every template in this repo that emits a single-source region now carries the meta tag: rpk command pages (`command.hbs`), Connect connector pages (`connector.hbs`), and both Bloblang overview pages. `connector.hbs` also stops emitting the TODO placeholder as the description *value* when a connector has none — a stub would inherit that into its meta tag.

## Consumer-side plan (tracked in [DOC-2414](https://redpandadata.atlassian.net/browse/DOC-2414))

- **adp-docs: nothing to do.** `reconcileStubs` now upgrades existing stubs in place when the partial has gained the region (idempotent, reported as `upgraded`), so the sync workflow self-heals.
- **cloud-docs: one-time scripted conversion PR** after this releases and upstream pages regenerate — no sync workflow exists there. Sequencing and verification steps are in the ticket.

## Validation

958 tests pass (4 new). The Antora experiment is reproducible: three-page site, production UI bundle, results as tabled above.

[DOC-2414]: https://redpandadata.atlassian.net/browse/DOC-2414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Jira

Part of [DOC-2414](https://redpandadata.atlassian.net/browse/DOC-2414) (enabler: the ticket itself is resolved by the cloud-docs conversion this unlocks).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,7 @@ mcp-usage-stats.json
 # Development rpk JSON files
 docs-data/rpk-vdev.json
 docs-data/rpk-vlocal.json
+
+# Property extractor working data (cloned/built at run time)
+tools/property-extractor/tmp/
+tools/property-extractor/tree-sitter/tree-sitter-cpp/
--- a/__tests__/tools/rpcn-connector-summaries.test.js
+++ b/__tests__/tools/rpcn-connector-summaries.test.js
@@ -400,6 +400,18 @@ describe('backfillPageDescriptions - self-healing page headers', () => {
     fs.rmSync(root, { recursive: true, force: true });
   });
 
+  test('maps the rate-limits data key to the rate_limits pages directory', () => {
+    const root = fs.mkdtempSync(path.join(os.tmpdir(), 'backfill-rl-'));
+    fs.mkdirSync(path.join(root, 'rate_limits'), { recursive: true });
+    fs.writeFileSync(path.join(root, 'rate_limits', 'local.adoc'),
+      '= local\n// tag::single-source[]\n:type: rate_limit\n\nBody.\n');
+    const result = backfillPageDescriptions({
+      'rate-limits': [{ name: 'local', summary: 'A simple X every Y rate limit.' }],
+    }, { pagesRoot: root });
+    expect(result.backfilled).toEqual(['rate_limits/local']);
+    fs.rmSync(root, { recursive: true, force: true });
+  });
+
   test('dry run reports without writing', () => {
     const root = fs.mkdtempSync(path.join(os.tmpdir(), 'backfill-dry-'));
     fs.mkdirSync(path.join(root, 'caches'), { recursive: true });
--- a/__tests__/tools/rpcn-connector-summaries.test.js
+++ b/__tests__/tools/rpcn-connector-summaries.test.js
@@ -340,3 +340,60 @@ describe('buildChangedDefaultsTable - whats-new changed default links', () => {
     expect(table).toContain('|batching.byte_size\n');
   });
 });
+
+describe('backfillPageDescriptions - self-healing page headers', () => {
+  const fs = require('fs');
+  const path = require('path');
+  const os = require('os');
+  const { backfillPageDescriptions } = require('../../tools/redpanda-connect/generate-rpcn-connector-docs');
+
+  const data = {
+    caches: [
+      { name: 'multilevel', summary: 'Combines multiple caches as levels,\nacross them.' },
+      { name: 'has_desc', summary: 'Should not be used.' },
+      { name: 'no_summary', summary: '' },
+      { name: 'no_page', summary: 'Page does not exist.' },
+    ],
+    'bloblang-functions': [{ name: 'not_a_page_family', summary: 'Ignored.' }],
+  };
+
+  test('inserts a tagged description into headers that lack one', () => {
+    const root = fs.mkdtempSync(path.join(os.tmpdir(), 'backfill-'));
+    fs.mkdirSync(path.join(root, 'caches'), { recursive: true });
+    fs.writeFileSync(path.join(root, 'caches', 'multilevel.adoc'),
+      '= multilevel\n// tag::single-source[]\n:type: cache\n:status: stable\n\nBody.\n// end::single-source[]\n');
+    fs.writeFileSync(path.join(root, 'caches', 'has_desc.adoc'),
+      '= has_desc\n// tag::single-source[]\n:description: Already here.\n\nBody.\n// end::single-source[]\n');
+    fs.writeFileSync(path.join(root, 'caches', 'no_summary.adoc'),
+      '= no_summary\n// tag::single-source[]\n:type: cache\n\nBody.\n// end::single-source[]\n');
+
+    const result = backfillPageDescriptions(data, { pagesRoot: root });
+    expect(result.backfilled).toEqual(['caches/multilevel']);
+    expect(result.skippedNoSummary).toEqual(['caches/no_summary']);
+
+    const lines = fs.readFileSync(path.join(root, 'caches', 'multilevel.adoc'), 'utf8').split('\n');
+    // Inserted at the end of the header (before the first blank line),
+    // summary flattened to one line, wrapped in the meta tag region.
+    expect(lines[4]).toBe('// tag::meta[]');
+    expect(lines[5]).toBe(':description: Combines multiple caches as levels, across them.');
+    expect(lines[6]).toBe('// end::meta[]');
+    expect(lines[7]).toBe('');
+    // Pages that already have one are untouched
+    expect(fs.readFileSync(path.join(root, 'caches', 'has_desc.adoc'), 'utf8')).toContain(':description: Already here.');
+
+    // Idempotent: second run backfills nothing
+    expect(backfillPageDescriptions(data, { pagesRoot: root }).backfilled).toEqual([]);
+    fs.rmSync(root, { recursive: true, force: true });
+  });
+
+  test('dry run reports without writing', () => {
+    const root = fs.mkdtempSync(path.join(os.tmpdir(), 'backfill-dry-'));
+    fs.mkdirSync(path.join(root, 'caches'), { recursive: true });
+    const page = '= multilevel\n// tag::single-source[]\n:type: cache\n\nBody.\n';
+    fs.writeFileSync(path.join(root, 'caches', 'multilevel.adoc'), page);
+    const result = backfillPageDescriptions(data, { pagesRoot: root, dryRun: true });
+    expect(result.backfilled).toEqual(['caches/multilevel']);
+    expect(fs.readFileSync(path.join(root, 'caches', 'multilevel.adoc'), 'utf8')).toBe(page);
+    fs.rmSync(root, { recursive: true, force: true });
+  });
+});
--- a/__tests__/tools/rpcn-connector-summaries.test.js
+++ b/__tests__/tools/rpcn-connector-summaries.test.js
@@ -400,6 +400,26 @@ describe('backfillPageDescriptions - self-healing page headers', () => {
     fs.rmSync(root, { recursive: true, force: true });
   });
 
+  test('flattens bare URL macros and internal xref shorthand out of the meta text', () => {
+    const root = fs.mkdtempSync(path.join(os.tmpdir(), 'backfill-url-'));
+    fs.mkdirSync(path.join(root, 'tracers'), { recursive: true });
+    fs.writeFileSync(path.join(root, 'tracers', 'open_telemetry_collector.adoc'),
+      '= open_telemetry_collector\n// tag::single-source[]\n:type: tracer\n\nBody.\n');
+    fs.mkdirSync(path.join(root, 'outputs'), { recursive: true });
+    fs.writeFileSync(path.join(root, 'outputs', 'broker.adoc'),
+      '= broker\n// tag::single-source[]\n:type: output\n\nBody.\n');
+    const result = backfillPageDescriptions({
+      tracers: [{ name: 'open_telemetry_collector', summary: 'Send tracing events to an https://opentelemetry.io/docs/collector/[Open Telemetry collector^].' }],
+      outputs: [{ name: 'broker', summary: 'Allows you to route messages to multiple child outputs using a range of brokering <<patterns>>.' }],
+    }, { pagesRoot: root });
+    expect(result.backfilled).toEqual(['tracers/open_telemetry_collector', 'outputs/broker']);
+    expect(fs.readFileSync(path.join(root, 'tracers', 'open_telemetry_collector.adoc'), 'utf8'))
+      .toContain(':description: Send tracing events to an Open Telemetry collector.');
+    expect(fs.readFileSync(path.join(root, 'outputs', 'broker.adoc'), 'utf8'))
+      .toContain(':description: Allows you to route messages to multiple child outputs using a range of brokering patterns.');
+    fs.rmSync(root, { recursive: true, force: true });
+  });
+
   test('maps the rate-limits data key to the rate_limits pages directory', () => {
     const root = fs.mkdtempSync(path.join(os.tmpdir(), 'backfill-rl-'));
     fs.mkdirSync(path.join(root, 'rate_limits'), { recursive: true });
--- a/__tests__/tools/rpcn-connector-summaries.test.js
+++ b/__tests__/tools/rpcn-connector-summaries.test.js
@@ -386,6 +386,20 @@ describe('backfillPageDescriptions - self-healing page headers', () => {
     fs.rmSync(root, { recursive: true, force: true });
   });
 
+  test('flattens AsciiDoc markup out of the meta text', () => {
+    const root = fs.mkdtempSync(path.join(os.tmpdir(), 'backfill-md-'));
+    fs.mkdirSync(path.join(root, 'processors'), { recursive: true });
+    fs.writeFileSync(path.join(root, 'processors', 'workflow.adoc'),
+      '= workflow\n// tag::single-source[]\n:type: processor\n\nBody.\n');
+    const result = backfillPageDescriptions({
+      processors: [{ name: 'workflow', summary: 'Executes a topology of xref:components:processors/branch.adoc[`branch` processors], in parallel.' }],
+    }, { pagesRoot: root });
+    expect(result.backfilled).toEqual(['processors/workflow']);
+    const page = fs.readFileSync(path.join(root, 'processors', 'workflow.adoc'), 'utf8');
+    expect(page).toContain(':description: Executes a topology of branch processors, in parallel.');
+    fs.rmSync(root, { recursive: true, force: true });
+  });
+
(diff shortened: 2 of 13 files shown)


JakeSCahill:
Upgraded from "duplicate, but tagged" to **single copy**: a follow-up experiment confirmed comment lines are legal inside an AsciiDoc header, so the `tag::meta[]` markers now wrap the page's one existing header `:description:` — no duplicate inside the single-source region at all. Verified in the same Antora-with-production-UI build: the source page keeps its own meta description, the stub inherits the identical line, body intact, zero warnings.

This fully retires the description duplication introduced in #225. One attribute serves the page itself and every stub. 957 tests green.

JakeSCahill:
@micheleRP both flags addressed:

1. **Precedence corrected** — re-verified in the Antora test build: later entry wins in both directions (include-then-literal → literal wins, literal-then-include → include wins). The in-place upgrader now strips the backfilled literal when it inserts the include (7d test added mirroring your finding), scoped inherently to generated reference stubs since the reconciler never touches prose stubs. PR body and DOC-2414 step 4 corrected — thanks for catching an inverted claim I'd already propagated to three places.
2. **Fallback at the #245 rebase** — agreed and noted in the body: `hasMetaTag ? dynamic include : literal copy`, so stubs created against pre-5.4 partials keep #245's literal instead of regressing to none. I'll implement exactly that shape when I rebase this over #245.

micheleRP (dismissed):
The core mechanism is correct, and I verified the AsciiDoc reasoning independently rather than taking the experiment on trust: with a header-only parse the reader stops at the first blank line, so an `include::` below it never contributes attributes — and an include on the line directly after the doctitle has its first line consumed as the **author line**, which is precisely why the no-blank-line variant destroys the body. Your table matches standard AsciiDoc header parsing exactly. The two-include shape is the right answer, and the `hasMetaTag` guard against tag-not-found warnings is the detail that makes graceful adoption actually work.

Requesting changes because the PR carries three separate changes beyond the stated one, and the largest of them dominates the PR's real-world impact without being reported anywhere.

## 1. `backfillPageDescriptions` rewrites roughly 290 published pages, unflagged and unreported

Measured against `rp-connect-docs` `main`: **31 of 326** connector pages carry a `:description:` at all. So the first `generate rpcn-connector-docs` run after this ships inserts a three-line meta block into roughly **290 hand-maintained published pages**, inside the ordinary auto-docs PR. That's the dominant effect of a PR titled "let single-source stubs inherit descriptions," and:

- **It runs unconditionally.** No CLI flag. The function supports `dryRun`, but nothing wires it through from the handler, so there's no way to preview it.
- **Nothing consumes the result.** `descriptionBackfill` is returned from `generateRpcnConnectorDocs` but no handler reads it, so unlike #227's `descriptionReports` and #236's `lostSectionWarnings` it never reaches the PR summary body. The reviewer facing a ~290-file diff gets one `console.log` line in a collapsed workflow log.
- **It's a new category of write.** This generator previously only wrote partials; it now edits `modules/components/pages/`. That's a real expansion of what "regenerate connector docs" means, and it deserves to be stated on the tin.

Suggest: surface the count and the `skippedNoSummary` list in the PR summary body, wire `dryRun` through, and consider landing the backfill as its own PR so the ~290-file diff can be reviewed as the thing it is rather than as a side effect.

## 2. Connectors with neither summary nor description emit no meta region at all

Dropping the TODO fallback is a genuine improvement — I confirmed two published pages currently ship `:description: // TODO: Missing description. Add a description in the overrides file or source data.` as their live meta description (`processors/string_split`, `tracers/open_telemetry_collector`). Good catch.

The consequence is that for those connectors the `meta` tag doesn't exist, so any stub including `[tag=meta]` gets an Antora tag-not-found warning. Those are exactly the 234 allowlisted stubs. So the DOC-2414 conversion script has to check for the tag per page the way `hasMetaTag` does for rpk — it can't just add the include everywhere. The PR body's "no help for the 234 allowlisted Connect stubs" reads as neutral when it's actually a hard constraint on the conversion.

Same issue for backfilled pages whose first blank line falls *above* `// tag::single-source[]`: the inserted meta block lands outside the tagged region, so the page has a description but `tag=meta` still won't resolve for a consumer. Worth having the backfill verify placement rather than assume it.

## 3. The PR body misdescribes the `command.hbs` change (minor)

The body says it "nests a `tag::meta[]` region around the in-tag `:description:`", and rests the safety argument on "every existing `tag=single-source` consumer renders byte-identically."

The diff actually wraps the **header** description and **deletes** the in-tag copy, so the single-source region's content did change. Rendered output is still identical, because a body-position attribute entry renders nothing — so there's no defect here. But the stated rationale is accurate for `connector.hbs` and the two Bloblang templates (where the region genuinely is nested) and not for `command.hbs`. Worth correcting, since that argument is what a reviewer leans on to skip checking consumers.

## 4. `dryRun` skips the upgrade path entirely (minor)

`if (partial.hasMetaTag && !dryRun)` means a dry run reports `upgraded: []`. The in-place stub rewrite is the one action here that mutates existing files, so it's the one you'd most want to preview. Separating the decision from the write would fix it.

## 5. The summary flattener misses the markup that's actually present (minor)

The flattener handles `xref:` / `link:` and backticks, but not the bare-URL macro form — which is what's in the data:

```
:description: A Kafka cache using the https://github.com/twmb/franz-go[Franz Kafka client library^].
:description: Send metrics to an https://opentelemetry.io/docs/collector/[Open Telemetry collector^].
```

Five descriptions already ship raw markup today, and the backfill will add more of the same shape. There are now three flatteners in flight — `connector.hbs` (none), #227's `summaryAttribute` (whitespace only), and this one — which should probably converge on one helper.

## 6. Two unrelated changes are bundled in (minor, scope)

- **`mergeOverrides` + `'summary'`** is a real bug fix. I confirmed five component-level summary overrides in `rp-connect-docs/docs-data/overrides.json` that are silently dropped today: zmq4 input, zmq4 output, ffi, kafka cache, otlp tracer. I also confirmed it's **safe** — example-level `summary` keys are handled by the `key === 'examples'` branch, which `continue`s before reaching `scalarKeys`, so nested example summaries are unaffected. It does change five pages' meta descriptions, and the kafka-cache one is an instance of item 5, so it deserves its own line in the body rather than arriving as a drive-by.
- **The `.gitignore` property-extractor entries** are unrelated to this PR entirely.

## Nits

- Created stubs get the meta include *after* the attribute lines; upgraded stubs get it *before* them (`replace(/^(= .+)$/m, ...)` inserts directly under the title). Both are valid header positions, but the inconsistency will puzzle the next reader.
- `.replace(/^(= .+)$/m, \`$1\n${metaInclude}\`)` — `metaInclude` contains `partial$rpk-ai`, and `$` is pattern-significant in a string replacement. `$r` isn't a special token so it survives today, but a function replacer removes the footgun for good.
- `upgraded_` with a trailing underscore, renamed to `upgraded` in the return object. `upgradedFiles` would avoid both.
- The Antora experiment is load-bearing for the entire design, but it lives only in the PR body — the tests assert emitted *text shape*, not that Antora actually produces the meta tag. Worth putting the reproduction somewhere durable so the next person can re-run it.

## Relationship to #245

Both PRs rewrite the same two sites (`partials.push({ file, title })` in `readPartialTitles`, and the `renderStub({...})` signature), so a conflict is guaranteed. Given how small #245 is, folding it in here may beat rebasing. The `hasMetaTag ? dynamic include : literal copy` composition is right — #245's literal copy stays the correct fallback for pre-5.4 partials.

## Version bump needed

This changes tool code (`generate-plugin-stubs.js`, `generate-rpcn-connector-docs.js`, three templates) with no `package.json` bump, and npm is at 5.3.5. Per the repo's publish rule, "rides whichever release goes out next" only works if a later PR bumps — otherwise these changes sit unpublished. #244 claims 5.3.6, #245 5.3.7, #227 5.4.0; this needs a version assigned rather than inherited.

JakeSCahill:
Rebased onto main now that #245 and #236 have merged, implementing the composition agreed on #245:

- `readPartialTitles` now extracts **both** signals: the literal header `:description:` and `hasMetaTag`.
- `renderStub` composes them: `hasMetaTag ? [tag=meta] include : (description ? literal copy : nothing)` — the dynamic include for partials generated with doc-tools >= 5.4, #245's literal copy as the fallback for older partials. The description mechanism (either form) is always the first header line after the title, per the ordering convention settled on #245, so created and upgraded stubs look alike.
- This also fixes a latent gap on this branch: `renderStub`'s body referenced `description` but the signature had dropped it, so the literal fallback could never fire (the no-meta test fixture had no description, which is why tests stayed green).
- Second conflict site was #236's `lostSectionWarnings` in the rpcn generator's return object — kept both fields.
- Version bumped to 5.3.8 (5.3.7 published with #245; #227 has claimed 5.4.0).

Full suite green after rebase: 1008 tests.

micheleRP (commented):
## Re-review after the rebase

**Overall assessment:** The core mechanism is sound and the rebase honored every agreement from the earlier (now-dismissed) review — but the new in-place upgrade path has one real safety gap, and its results are invisible in both dry-run and the sync workflow's PR summary, which means the first real run after release will rewrite every existing adp-docs stub while reporting "No changes."

### Verified as addressed from the earlier review

The precedence inversion is corrected everywhere with a mirroring test; the #245 composition (`hasMetaTag ? dynamic include : literal copy`) is implemented exactly as agreed — including fixing a latent gap where `renderStub`'s literal fallback could never fire; created and upgraded stubs now share the same header ordering; the version is bumped to 5.3.8 (npm is at 5.3.7); and the one-time ~290-page connector backfill now lands as its own reviewable PR (rp-connect-docs#478), which was the main structural ask. CI is fully green.

### DOC-2414 alignment

The PR is the stated enabler and matches the ticket's design, including the corrected precedence rule. Three mismatches:

- Ticket step 3 says adp-docs upgrades are "reported as `upgraded`" — true only at the function-return level; the CLI never surfaces it (critical issue 2 below).
- The ticket promises "Leave prose stubs untouched" — the upgrade path lacks the guard that would guarantee that (critical issue 1).
- Ticket step 1 still says "no bump — rides the next release"; the rebase now bumps to 5.3.8. Stale ticket text, worth a one-line edit.

### Critical issues (must fix)

1. **[`tools/rpk-docs/generate-plugin-stubs.js:235`] The upgrade path can rewrite hand-maintained pages.** The delete pass carefully guards with `managedStubRe` ("Pages that do not match the managed-stub shape are never deleted: they may be hand-written"), but the new upgrade branch has no such check — it fires for *any* `.adoc` in the stub directory whose filename matches a partial. A hand-written prose page named like a partial would get its first `:description:` deleted and a meta include inserted. The comment claiming "Hand-maintained prose stubs never pass through this reconciler" is false — this branch is exactly where they pass through, and it contradicts DOC-2414's "leave prose stubs untouched" guarantee. Exposure in adp-docs today is probably low (stubs there are all managed), but the fix is one condition: only upgrade when `managedStubRe.test(stubContent)`.
   - **Fix:** Gate the upgrade on the managed-stub shape, and correct the comment.

2. **[`bin/doc-tools.js:1302`] `upgraded` is invisible everywhere, and the "No changes" message is wrong on the first real run.** The console line prints only created/deleted/nav; the `--summary-file` markdown (which becomes the sync PR body) has no upgraded section; and "No changes: stubs and nav already match the partials." triggers whenever `created + deleted === 0 && !navUpdated`. The first sync after 5.3.8 releases and the rpk partials regenerate will upgrade **every existing stub in place** — created=0, deleted=0 — so the PR body will assert "No changes" over a diff that rewrites the whole stub set. That's guaranteed, not an edge case.
   - **Fix:** Add `upgraded` to the console line, the summary markdown, and the no-changes condition.

### Suggestions (should consider)

3. **[`generate-plugin-stubs.js:235`] `--dry-run` skips the upgrade path entirely** (`partial.hasMetaTag && !dryRun`), so a dry run reports `upgraded: []` for the one action that mutates existing files. Compute the decision always; gate only the write. (Carried over from the earlier review, still unaddressed.)

4. **[`generate-rpcn-connector-docs.js:682`] `descriptionBackfill` is returned but never consumed.** The handler plumbs `lostSectionWarnings` into the PR summary (the #236 precedent) but drops `descriptionBackfill`; there's also no CLI dry-run wiring. With #478 carrying the one-time diff this is now about future drift-heals, but those will still arrive silently inside auto-docs PRs. Wire it into `printPRSummary` alongside the lost-section warnings.

5. **[`generate-rpcn-connector-docs.js:746`] The summary flattener still misses the bare-URL macro form** (`https://...[label^]`) — the shape actually present in the data (franz-go, otel), now more relevant because the `summary` override fix activates the kafka-cache override containing exactly that markup, and `connector.hbs` templates `{{{summary}}}` raw. (Carried over, still unaddressed.)

6. **[`connector.hbs:4`] Pre-existing, but now it defeats this PR's purpose:** `{{#if support}}:support: ...{{/if}}` leaves a blank line when `support` is falsy (verified with the repo's Handlebars), and `support` is only set for components found in the CSV metadata. For such components the header ends at that blank line and the meta block lands in body position — the page ships the generic description (stubs still inherit fine, since include-tag filtering is file-wide). Cheap to fix while in the template.

7. **PR body is stale/incomplete:** (a) it still says command.hbs "nests a `tag::meta[]` region around the in-tag `:description:`" with the byte-identical-via-tag-stripping rationale — the actual diff wraps the *header* copy and deletes the in-tag copy (render-identical for a different reason: a removed body attribute entry renders nothing); (b) `backfillPageDescriptions`, the `mergeOverrides` fix, and the helm relocation aren't mentioned in the Changes list at all; (c) "No version bump" is stale; (d) "234 allowlisted Connect stubs" is stale versus the ticket's collapse to 7.

8. **Minor code nits:** comments say "doc-tools >= 5.4" twice in `generate-plugin-stubs.js` while this ships as 5.3.8; `reconcileStubs`' `@returns` JSDoc omits `upgraded`; `mergeOverrides`' docstring wasn't updated for `summary`; the `$`-token replacement footgun is now in two places (`metaInclude` in the upgrader, `descMatch[0]` in the helm relocation); the helm relocation's file-wide `\n{3,}` collapse could alter blank lines inside literal blocks of the converted README; the `.gitignore` property-extractor entries remain unrelated to this PR; and if #227 (which claims 5.4.0) merges and publishes first, 5.3.8 will never publish — whoever merges second re-bumps, per repo rules.

### What works well

- The empirically verified two-include mechanism, with the experiment's conclusions now encoded in durable test comments.
- Dropping the TODO-fallback as a description *value* — a real bug affecting two live pages.
- The `mergeOverrides` `summary` fix with a regression test explaining why the key earned its place.
- Test quality throughout: idempotency assertions, the no-warn guard for pre-5.4 partials, and the literal-strip test that mirrors the precedence finding.
- Restructuring the mass backfill into rp-connect-docs#478 so the big diff is reviewed as the thing it is.

micheleRP (approved):
Re-reviewed just the changes since my last review (`933851a7` plus the merge from main).

LGTM. I ran the test suite on the PR head (30/30 passing) and checked the new regexes against real connector data (`connect-4.103.1.json` and `overrides.json`): all 31 summaries with bare URL macros and all 4 with `<<...>>` refs flatten cleanly, including the leading-space label in `<<codecs, structured data>>`.

A few non-blocking notes:

1. **The caret strip is broader than its comment says.** `.replace(/\^/g, '')` removes every caret in the summary, not just the `[label^]` link caret. It's safe against current data (all 24 carets in top-level component summaries are link-label carets; the regex-caret summaries like `` `^foo.*` `` only exist at the `examples[]` level, which doesn't feed this path). But a future summary documenting a regex would be silently mangled. A tighter option: fold the caret into the label captures (`\[([^\]]*?)\^?\]`) on the two link regexes and drop the global strip.
2. **Test gap:** the new test covers `https://...[label^]` and `<<patterns>>` but not the two-part `<<anchor,label>>` form, which occurs in real data (e.g. the `awk` processor's `<<awk-functions,custom functions>>`). The regex handles it — just untested.
3. **Version check:** the merge resolution sets the package version to 5.10.0 while main and npm are both at 5.7.0. If 5.8/5.9 are claimed by other open PRs, all good — just confirming it's deliberate and not a typo for 5.8.0.

Follow-up idea (not this PR): the flattening only exists in `backfillPageDescriptions` — `connector.hbs` still emits `:description: {{{summary}}}` raw, so newly drafted pages ship unflattened markup in their meta descriptions while backfilled pages come out clean (67 summaries in current data carry markup; 4 contain embedded newlines that would break the header attribute). Extracting the flattener into a shared Handlebars helper would close that gap. Happy to open an issue for it.

PR #254, opened 2026-08-08
Add the prop macro for opt-in, validated property references

JakeSCahill:
## In plain English

The docs used to guess which bits of code were configuration properties, and guessed wrong for property names that are ordinary words (`admin`, `rack`, `brokers`, `retries`, `superusers`) -- readers got Redpanda property tooltips on words that meant something else entirely, like the `admin` audit-logging value in the Helm chart. This PR adds a `prop:` macro so writers say explicitly "this is a property," and the build checks every name against the published property data, so a typo can't ship. When a property should link to its reference page, the build finds the right page by reading the docs themselves rather than trusting a hardcoded rule -- so links stay correct per doc set (streaming, cloud, and so on) and keep working if the property reference is ever split into different pages. It also retires the old `config_ref` macro, which never validated names and stamped a tiered-storage Helm path onto *every* property on Kubernetes pages, mislabeling the non-tiered ones on the live site. A migration command converts existing content and refuses to touch anything it can't verify.

Full explainer for the team: `macros/PROPERTY_AND_ENTERPRISE_REFERENCES.adoc` (added here).

## Description

Property tooltips currently decorate **every** inline code element whose text matches a property name. The published JSON has exactly 5 separator-free names — `admin`, `brokers`, `rack`, `retries`, `superusers` — all common words that mean other things in Helm/Kubernetes/Connect contexts (`admin` is also an audit-logging value in the Helm chart), so readers get wrong tooltips. Same ontology hole the enterprise macro closed: implicit matching on ambiguous strings.

### prop: inline macro

`prop:cloud_storage_enabled[]` emits `<code class="property-ref" data-property-name=...>` for the docs UI to decorate — marking becomes **opt-in**. Targets are validated against the published property reference (the newest `redpanda-properties-<tag>.json` attachment already in the content catalog — the exact data the tooltips fetch at runtime), controlled by `property-validate` (`warn` default with file path + did-you-mean, `error`, `off`). Graceful degradation with a single warning when no JSON is in the build.

### config_ref blending

`prop:name[link=true]` links the property to its reference page via dynamic discovery (see below), with `page=` override. **`config_ref` is fully superseded and deprecated.** Investigation showed its Kubernetes behavior — a blanket `storage.tiered.config.` prefix on every property on `env-kubernetes` pages — mislabels non-tiered properties (the live k-tiered-storage page renders `storage.tiered.config.log_segment_size`, while the docs' own guidance says `config.cluster` is the home for all cluster properties). The replacement is `helm-path=auto`: on `env-kubernetes` renders, the property displays as its per-property correct Helm values path (`storage.tiered.config.*` only for tiered storage properties, `config.node.*` for broker properties, `config.cluster.*` otherwise) — a data-driven fix, not a hardcoded prefix. The codemod gains `--config-refs` to convert all call sites (path argument dropped in favor of discovery, verified disjoint across the property partials; payloads repeating the name are dropped, differing payloads become `text=`). `config_ref` stays registered only because the generated property partials still emit it — a property-extractor follow-up.

### Migration codemod

`doc-tools generate migrate-property-refs` (dry run unless `--write`): converts `\`name\`` → `prop:name[]` only for names that exist in the published JSON **and contain a separator** — the 5 ambiguous names are never auto-converted. Skips autogenerated content (rpk pages, k-crd, Helm specs, property partials), code/literal/comment/fenced blocks, headings, attribute entries, block titles, and other macros' `[]` payloads.

Dry run against the real docs repo: **647 conversions across 86 hand-written files** (825 → 452 files scanned after excluding generated content), ambiguous names left alone.

## Companion PRs

- docs-ui: switches `19-property-tooltips.js` from text-matching to decorating only `[data-property-name]` elements (stacked on #413's branch).
- docs: codemod run + macro registration (stacked on the registry-adoption branch).

## Validation

- 29 new tests (macro conversion incl. dotted topic properties, newest-JSON selection, validation modes, caching, degradation; codemod line/block/classification behavior). Full suite: 1082.
- Live dry run above; sample diffs eyeballed (table cells and prose convert, code blocks and macro payloads untouched).

## Version

Bumps to 5.15.0 (re-laddered; see #253). **Stack**: #249 (5.12.0) ← #253 (5.13.0) ← #251 (5.14.0) ← this PR (5.15.0). Merge in order and versions work out.

## Update: build-time doc-URL stamping (Aug 11)

`68b7dd83` — live review on the demo site showed tooltip documentation links were guessed client-side from `config_scope` and were wrong three ways on cloud pages (nonexistent topic-properties page, wrong page for object-storage properties, invented anchor ids). Discovery now runs for every `prop:` marker — not just `link=true` — and stamps `data-doc-url` with the documenting page's published URL plus the real Asciidoctor anchor (dots→hyphens, underscores kept). `link=true` xref fragments use the same anchor rule (dotted names previously pointed at nonexistent fragments). The undocumented-property warning fires only when a link was requested. docs-ui#415 prefers the stamp.


kbatuigas (approved):
R.I.P., `config_ref` 🫗

PR #257, opened 2026-08-12
fix(redpanda-connect): avoid rpk connect upgrade's version-regex bug in report-delta

JakeSCahill:
## Why

`getRpkConnectVersion()` (used when `doc-tools generate rpcn-connector-docs --fetch-connectors` runs without `--connect-version`, i.e. the normal CI path) called `rpk connect upgrade` to make sure the plugin was current before reading its version.

`rpk connect upgrade` parses the *currently installed* Connect version through `redpanda.VersionFromString` (`rpk/pkg/redpanda/version.go`) to compare it against latest. That parser's regex caps every version segment at two digits:

```go
regexp.MustCompile(`^v?(\d{1,2})\.(\d{1,2})\.(\d{1,2})(?:\s|-rc\d{1,2}|-dev|-nightly|$)`)
```

So it throws for any installed Connect version >= 4.100.0 — every currently shipping version. This is the same bug class as [CON-529](https://redpandadata.atlassian.net/browse/CON-529) / [redpanda#31432](https://github.com/redpanda-data/redpanda/issues/31432) (fixed in [redpanda#31438](https://github.com/redpanda-data/redpanda/pull/31438)), but in a different function that fix didn't touch — confirmed still present and unfixed on rpk `dev` as of this writing. It's what broke the scheduled `Update RPCN Connector Docs` workflow in rp-connect-docs starting 2026-08-07 (`Cannot process version X - data unavailable`), even after CON-529's own fix landed and after switching to an rpk build (`v26.1.15`) that has it.

## What this does

Swaps the plugin-freshness check from `rpk connect upgrade` to `rpk connect install --force`. Both ensure the latest plugin is installed; `install`'s default version is the string `"latest"`, which its own validator short-circuits before ever reaching a version-parsing regex — so this path isn't exposed to the bug at all, on any rpk build (verified against both the pre-fix v26.2.1 and the post-fix v26.1.15).

Filed against rpk upstream separately; this is a workaround so our pipeline doesn't have to wait on another rpk release.

## Test plan

- [x] Full test suite passes (`npx jest`, 1020/1020)
- [x] Manually verified `rpk connect install --force` succeeds on both v26.2.1 (pre-fix) and v26.1.15 (post-fix) rpk builds, whereas `rpk connect upgrade` fails on both once a >= 4.100.0 version is installed
- [x] Used this patched version to successfully regenerate rp-connect-docs connector docs locally (redpanda-data/rp-connect-docs#489)

[CON-529]: https://redpandadata.atlassian.net/browse/CON-529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tools/redpanda-connect/report-delta.js
+++ b/tools/redpanda-connect/report-delta.js
@@ -424,8 +424,15 @@ function buildComponentMap(indexObj) {
 
 function getRpkConnectVersion() {
   try {
-    // Make sure the connect plugin is upgraded first (silent)
-    execSync('rpk connect upgrade', { stdio: 'ignore' });
+    // Ensure the connect plugin is current (silent). Deliberately "install
+    // --force", not "upgrade": upgrade's version comparison parses the
+    // currently-installed version through a regex capped at two digits per
+    // segment (redpanda.VersionFromString in rpk/pkg/redpanda/version.go),
+    // which throws for any Connect version >= 4.100.0 — the same bug class
+    // as CON-529, in a different function that its fix didn't touch, and
+    // still unfixed as of rpk dev. install's "latest" path (the default)
+    // skips that regex entirely, so it isn't exposed to the bug.
+    execSync('rpk connect install --force', { stdio: 'ignore' });
 
     // Now capture the --version output
     const raw = execSync('rpk connect --version', {


kbatuigas (approved):
Please see the finding, but I'll go ahead and approve:

> - **Medium — `install --force` silently replaces/shadows self-managed Connect installations.** The old `rpk connect upgrade` explicitly refuses to operate when Connect came from Homebrew or another package manager. `rpk connect install --force` skips that safeguard, writes a managed binary into `~/.local/bin`, and rpk prioritizes that location. A local `--fetch-connectors` run can therefore change which Connect binary the user runs afterward. Preserve the self-managed check or restrict the forced installation to CI. See [[changed line](https://github.com/redpanda-data/docs-extensions-and-macros/blob/17203bf3d6796ca52a8ecb0b9d39da0f8aefd83e/tools/redpanda-connect/report-delta.js#L437)](https://github.com/redpanda-data/docs-extensions-and-macros/blob/17203bf3d6796ca52a8ecb0b9d39da0f8aefd83e/tools/redpanda-connect/report-delta.js#L437) and rpk’s [[upgrade safeguard](https://github.com/redpanda-data/redpanda/blob/dev/src/go/rpk/pkg/cli/connect/upgrade.go)](https://github.com/redpanda-data/redpanda/blob/dev/src/go/rpk/pkg/cli/connect/upgrade.go).
> 
> Non-blocking: the PR says `rpk-overrides.schema.json` ships with the npm package, but `package.json` excludes `docs-data/`, and `npm pack --dry-run` confirms the schema is absent. The canonical docs repo has its own sibling schema, so current automation works; either adjust the claim or add the schema to the package whitelist.
> 
> Verification:
> 
> - Branch merges cleanly with current `main`.
> - Targeted tests: 21/21 passed.
> - Official [[CI matrix is green](https://github.com/redpanda-data/docs-extensions-and-macros/actions/runs/32147893881)](https://github.com/redpanda-data/docs-extensions-and-macros/actions/runs/32147893881).
> - The `install --force` workaround itself correctly bypasses the version parser bug.
> 
> Verdict: request changes for the self-managed-install regression.

JakeSCahill:
Thanks for the review — both findings addressed.

**Medium: `install --force` shadowing a self-managed Connect** — `d431bd57`

Confirmed the safeguard you pointed at: `upgrade.go` dies on `!connect.Managed`, and `install --force` has no equivalent. Took the "restrict to CI" option rather than reimplementing the `Managed` check, since detecting it from Node would mean parsing rpk's plugin listing and coupling us to that output format.

`shouldRefreshConnectPlugin()` now gates the refresh on `CI`, with `RPCN_FORCE_CONNECT_INSTALL` as an override in either direction. Locally the version is read from whatever Connect is already installed, so a Homebrew or hand-installed binary is never shadowed, and the failure path now says how to install a plugin instead of only reporting that the call failed. The comment at the call site records both constraints — the regex bug that forces `install`, and the safeguard that confines it to CI.

**Non-blocking: `rpk-overrides.schema.json` not in the tarball** — `6c9757de`

You're right, `npm pack --dry-run` had zero matches: `files` has no `docs-data/` entry. Since shipping it is the intent, I whitelisted the single file rather than softening the claim. `npm pack` now lists `27.3kB docs-data/rpk-overrides.schema.json`, with the rest of `docs-data/` still unpublished.

**Tests:** targeted suite 8/8 (was 3), green with `CI` unset and `CI=true` — the previous version was env-dependent, so it now pins the variable explicitly. Full suite 1027/1027. The new `outside CI` block is the regression guard: it asserts `rpk connect install` is never invoked and that `rpk connect --version` is the only command run.

PR #258, opened 2026-08-12
fix(redpanda-connect): recursively detect new fields nested under existing groups

JakeSCahill:
## Why

`rp-connect-docs#489` (a manual run of the connector docs generation) added two genuinely new fields — `sqs.zero_key_warn_interval` on `aws_s3` and `logminer.max_session_age` on `oracledb_cdc` — but neither showed up in `whats-new.adoc`'s "New field support" section or the PR's "Writer Action Items", even though both are confirmed absent from the prior release's connector data and present in the new one.

Root cause: `buildComponentMap()` only records **top-level** field names:

```js
const fieldNames = childArray.map(f => f.name)
```

Both new fields are nested one level inside a group that already existed (`sqs`, `logminer`). Since the group name itself didn't change, the diff saw the top-level field set as unchanged and reported zero new fields for both intermediate releases — this isn't specific to that PR; any field added under an existing nested group hits the same blind spot.

## What this does

- `flattenFields()` recursively walks each field's `children` and returns dotted paths (`sqs.zero_key_warn_interval`), so nested additions are visible to the same Set-based comparison that already worked for top-level fields. `buildComponentMap()` now also builds a `fieldMap` (path → raw field object) alongside the flat name list, which replaces the repeated `type === 'config' ? ... : ...` lookup branching duplicated ~8 times across `generateConnectorDiffJson` and `printDeltaReport` with a single `.get(path)` call.
- New fields now get an `introducedIn` stamped from the diff's own `opts.newVersion` when the source data carries no per-field version (it never does today — checked both real fields, neither has `introducedInVersion`/`version`). Field-level data still wins if it's ever populated.
- `buildFieldsTable` renders that as an "Introduced in" column, opt-in via `{ showIntroducedIn: true }` (only the "New field support" call site uses it — the deprecated-fields table doesn't need it).

## Test plan

- Added a test reproducing the exact bug shape (field added under an existing group) and confirming it's now detected.
- Added tests for the version-stamping fallback and the field-level-version-wins case.
- Added `buildFieldsTable` tests for the new opt-in column (present/absent, and the `-` fallback when a field has no version).
- Ran the fixed `generateConnectorDiffJson` against the real `connect-4.103.1.json` → `connect-4.104.0.json` data from rp-connect-docs#489: both previously-missed fields now appear, each correctly stamped `introducedIn: '4.104.0'`.
- Full suite: `npx jest` — 1037/1037 pass.

### DIFF
--- a/__tests__/tools/rpcn-connector-summaries.test.js
+++ b/__tests__/tools/rpcn-connector-summaries.test.js
@@ -156,6 +156,88 @@ describe('generateConnectorDiffJson - config component status and cloud support'
     expect(inDetails).toBe(true);
     expect(added.cloudSupported).toBe(true);
   });
+
+  test('detects a new field nested under an existing group, not just new top-level fields', () => {
+    // aws_s3's "sqs" group already exists in both versions; only a field
+    // nested inside it is new. A diff that only compares top-level field
+    // names would see "sqs" unchanged and report zero new fields, even
+    // though sqs.zero_key_warn_interval is genuinely new.
+    const oldIndex = {
+      inputs: [
+        {
+          name: 'aws_s3',
+          config: {
+            children: [
+              { name: 'bucket', description: 'Bucket name.' },
+              {
+                name: 'sqs',
+                children: [
+                  { name: 'url', description: 'Queue URL.' }
+                ]
+              }
+            ]
+          }
+        }
+      ]
+    };
+    const newIndex = {
+      inputs: [
+        {
+          name: 'aws_s3',
+          config: {
+            children: [
+              { name: 'bucket', description: 'Bucket name.' },
+              {
+                name: 'sqs',
+                children: [
+                  { name: 'url', description: 'Queue URL.' },
+                  { name: 'zero_key_warn_interval', description: 'Warn interval.' }
+                ]
+              }
+            ]
+          }
+        }
+      ]
+    };
+
+    const diff = generateConnectorDiffJson(oldIndex, newIndex, {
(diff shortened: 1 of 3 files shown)


kbatuigas (approved):
I'll approve, but sharing this finding just in case:

> 7. **[P3] PR #258 — nested array paths omit `[]`**  
>    [`[report-delta.js:376](https://github.com/redpanda-data/docs-extensions-and-macros/blob/f0b240c/tools/redpanda-connect/report-delta.js#L376)`](https://github.com/redpanda-data/docs-extensions-and-macros/blob/f0b240c/tools/redpanda-connect/report-delta.js#L376)
> 
>    `flattenFields()` uses raw field names, while rendered reference fields append `[]` for arrays. A nested addition is reported as `sasl.aws...` instead of `sasl[].aws...`. Share the renderer’s path-segment helper and add a nested-array regression test.

JakeSCahill:
@kbatuigas thanks for flagging — fixed in 11a1132. `flattenFields` is gone; both the diff (`report-delta.js`) and the renderer now walk field trees through the same `flattenConnectFields` helper (`tools/redpanda-connect/helpers/flattenConnectFields.js`, added in 108bf06) with `{ arrayMarker: true }`, so a nested addition under an array-typed group now reports as `sasl[].aws...`, matching the heading the reference page actually renders. Regression test covers a field added under an existing array group.

(Also addressed CodeRabbit's `byField`/`introducedIn` finding in a713f87/6e6f4ae — replied on that thread.)

PR #259, opened 2026-08-13
fix(ci): isolate the CGO network test and report download failures honestly

JakeSCahill:
Two fixes for the flake that failed #248 — a PR touching only helm-spec code, with no relation to CGO detection.

## 1. `test-all` re-included a test the config deliberately excludes

`jest.config.js` already ignores the CGO end-to-end test, because it downloads real release binaries:

```js
testPathIgnorePatterns: ['/node_modules/', '__tests__/tools/cgo-detection.test.js']
```

But a CLI `--testPathIgnorePatterns` **replaces** that array rather than adding to it. So `test-all` passing `"cli-contract.test.js|integration.test.js"` silently dropped both the cgo exclusion and `/node_modules/`.

Verified with `jest --listTests`:

| invocation | cgo test collected? |
|---|---|
| config only | no |
| config + old CLI flag | **yes** |
| config + new CLI flag | no |

This also explains why `test-tools` passed on the same commit — it globs `__tests__/tools/` but doesn't pass the flag, so the config exclusion held.

The ignore list now repeats the config's own entries. The test still runs where it's meant to: `test-cgo-detection.yml` and `npm run test:cgo`, both of which invoke it as a node script. Restoring `/node_modules/` is defensive — no dependency test files are collected today, but the config intended to exclude them.

## 2. A failed download reported itself as a corrupt archive

`curl -sL` has no `--fail`, so on an HTTP error it exits 0, writes the error body into `cloud.tar.gz`, and the run dies in `tar`:

```
gzip: stdin: not in gzip format
tar: Child returned status 1
```

That reads like a bad archive rather than a 404 or a rate limit, which is what made the original failure expensive to diagnose. Reproduced directly:

```
curl -sL  <missing asset>  -> exit 0, wrote 9 bytes of non-tarball
curl --fail <missing asset> -> exit 56, no file left behind
```

Changes:
- `--fail` so an HTTP error surfaces as an HTTP error
- `--retry 3 --retry-delay 2`, covering the 429 and 5xx statuses GitHub returns when throttling release downloads
- a checksum check against the `.sha256` published beside each asset, so a truncated download is named as such. Hashing uses node `crypto` rather than `sha256sum`/`shasum`, which differ between Linux and macOS.

## Verification

Against the real `v4.75.1` cloud asset (73MB):

1. download with `--fail --retry` succeeds
2. checksum matches the published `.sha256`
3. archive extracts cleanly
4. a truncated file is correctly reported as a checksum mismatch
5. a missing `.sha256` degrades to a warning rather than failing the run

Full suite green on a clean `npm ci`: 54 suites, 1198 tests.

Note: `#248` itself needed no change — I re-ran its failed job and it passed.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -287,8 +287,15 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      # A CLI --testPathIgnorePatterns REPLACES the array in jest.config.js, it
+      # does not add to it. So this list has to repeat the config's own entries:
+      # without /node_modules/ jest walks dependency test files, and without
+      # cgo-detection.test.js the network-dependent CGO end-to-end test runs
+      # here, where a GitHub download hiccup fails PRs that never touched it.
+      # That test has its own workflow (test-cgo-detection.yml) and npm script
+      # (npm run test:cgo), which run it as a node script.
       - name: Run all tests
-        run: npm test -- --testPathIgnorePatterns="cli-contract.test.js|integration.test.js" --maxWorkers=2
+        run: npm test -- --testPathIgnorePatterns="/node_modules/|cli-contract.test.js|integration.test.js|cgo-detection.test.js" --maxWorkers=2
 
   # Summary job that all other jobs depend on
   test-summary:
--- a/__tests__/tools/cgo-detection.test.js
+++ b/__tests__/tools/cgo-detection.test.js
@@ -14,6 +14,7 @@
 
 const fs = require('fs');
 const path = require('path');
+const crypto = require('crypto');
 const { execSync, spawnSync } = require('child_process');
 const os = require('os');
 
@@ -74,6 +75,41 @@ function getPlatformInfo() {
   };
 }
 
+/**
+ * Verify a downloaded asset against the .sha256 published beside it.
+ *
+ * Catches a truncated or substituted download before tar reports it as a
+ * generic archive error. Hashing in node keeps this working on both Linux
+ * (sha256sum) and macOS (shasum), which have different CLIs.
+ */
+function verifyChecksum(filePath, sha256Url, assetName) {
+  let published;
+  try {
+    published = execSync(`curl -sSL --fail --retry 3 --retry-delay 2 "${sha256Url}"`, {
+      stdio: ['pipe', 'pipe', 'pipe']
+    }).toString();
+  } catch (err) {
+    // No checksum published for this asset: keep going rather than fail the
+    // run over a missing side file, since the archive itself still has to
+    // extract and execute below.
+    log(`  Warning: no checksum available at ${sha256Url}, skipping verification`);
+    return;
+  }
+
+  // Format is "<hash>  <filename>"; take the first token.
+  const expected = published.trim().split(/\s+/)[0];
+  const actual = crypto
+    .createHash('sha256')
+    .update(fs.readFileSync(filePath))
+    .digest('hex');
+
+  if (expected && expected !== actual) {
+    throw new Error(
+      `checksum mismatch for ${assetName}\n  expected: ${expected}\n  actual:   ${actual}`
+    );
+  }
+}
+
 function downloadBinary(version, type) {
   const platform = getPlatformInfo();
   const binaryName = type === 'cgo' ? 'redpanda-connect-cgo' : 'redpanda-connect-cloud';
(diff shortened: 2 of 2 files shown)


JakeSCahill:
Good catch, and a fair hit — fixed in 99dc0c4f. The first pass had the same defect it was meant to remove: the catch-all treated every retrieval failure as "no checksum published" and continued, so a 5xx, a DNS failure, a timeout, or an empty 200 body all skipped verification while reporting success. That is precisely the silent-success behavior that made `curl -sL` misleading here in the first place.

Now only a confirmed HTTP 404 skips, since that genuinely means no `.sha256` was published beside the asset. Everything else fails:

| condition | before | now |
|---|---|---|
| HTTP 404 | skip | skip (intended) |
| HTTP 500 | skip silently | throws with status |
| transport error / DNS | skip silently | throws |
| empty or whitespace body | skip silently | throws |
| body not a 64-char hex SHA-256 | skip silently | throws |
| hash mismatch | throws | throws |

The status is available because `fetchWithStatus` uses `-w '%{http_code}'` and deliberately omits `--fail` — with `--fail`, 404 and 500 collapse into the same exit code, which is what made the two indistinguishable. Comparison is now case-insensitive too.

I also took the OpenGrep command-injection warning rather than dismissing it: the download, checksum fetch, and tar extraction now use `execFileSync` with argument arrays, so no URL or path passes through a shell. That also removes a latent quoting bug if the checkout path ever contains spaces.

Verified all seven branches, including a real 404 against the release and an unresolvable host: 404 skips with a warning, matching hash passes, and mismatch / HTTP 500 / empty body / HTML body / transport error all throw.

Not acted on: ast-grep's path-traversal warning on `fs.readFileSync(filePath)`. That path is built internally from `__dirname` and a pinned version string with no external input, so there is nothing to normalize.

micheleRP (approved):
## Review: isolate the CGO network test and harden downloads

Approving. This removes a real class of false CI failure and the reasoning is documented where it matters.

The core insight is right: `curl` without `--fail` exits 0 on an HTTP error, writes the error body into the `.tar.gz`, and the failure then surfaces from `tar` as "gzip: stdin: not in gzip format" — which reads like a corrupt archive rather than a 404 or a rate limit. Moving to `execFileSync` with argument arrays also takes the URL and paths out of shell interpretation, which is a nice secondary win.

### CodeRabbit's Major finding is already addressed

It asked to distinguish a missing checksum asset from a retrieval failure. Commit `99dc0c4` ("fail loudly when a checksum cannot be retrieved or parsed") landed at 15:38:27Z, five minutes after the 15:33:32Z comment, so it reviewed the initial commit only. The current code treats only a 404 as absent, throws on any other status, and validates the value against `SHA256_RE` before comparing. No action needed — flagging so it does not get re-raised.

The accompanying ast-grep path-traversal warning on `fs.readFileSync(filePath)` is noise; the path is constructed internally.

### Suggestions

**1. The ignore list now lives in two places.** The comment correctly explains that `--testPathIgnorePatterns` on the CLI *replaces* the array in `jest.config.js` rather than extending it — which is precisely why copying those entries into the workflow is a drift trap. Add an ignore entry to the config later and this workflow silently stops honoring it, with nothing to catch the divergence. A dedicated npm script (`test:ci`) or a jest project would keep the list in one place.

**2. A missing checksum is treated as acceptable.** `verifyChecksum` logs and returns on a 404. Release assets should always publish a `.sha256`, so a 404 in CI is arguably itself a failure worth surfacing. The behavior is deliberate and documented, so this is only to make the choice explicit rather than inherited.

### What works well

- Every comment names the failure mode it prevents rather than restating the code. That is why this diff reviews quickly despite touching CI.
- Hashing in Node rather than shelling out to `sha256sum`/`shasum` correctly sidesteps the Linux/macOS CLI difference.
- Deliberately omitting `--fail` in `fetchWithStatus` so the status code stays available to the caller is the right call, and the comment says why.
- Moving the network-dependent test to its own workflow is the right structural fix — a test that depends on a third-party download should not be able to fail PRs that never touched it.

JakeSCahill:
@coderabbitai The checksum-retrieval finding was addressed in 99dc0c4 ("fail loudly when a checksum cannot be retrieved or parsed"): only an HTTP 404 is treated as "no checksum published" and skips verification with a log line; any other status, a transport error, an empty body, or a body that is not a 64-character SHA-256 now throws. The latest push also single-sources the CI test exclusions (jest.unit.config.js) so the workflow no longer duplicates jest.config.js's ignore list.

PR #260, opened 2026-08-15
fix(validate): honor --skip-connect for antora.yml and authenticate private fetches

JakeSCahill:
## Problem

`doc-tools validate enterprise-features` fails on every run, which is why the `validate` check is red on redpanda-data/docs#1895 and #1896:

```
ERROR [fetch] Could not load rp-connect-docs antora.yml
(https://raw.githubusercontent.com/redpanda-data/rp-connect-docs/main/antora.yml): 404 Not Found
```

Two separate causes behind it.

**1. `--skip-connect` was only half-implemented.** It gated the connect `info.csv` fetch but not the `rp-connect-docs` `antora.yml` fetch. That value feeds only `checkConnect` (`tools/enterprise-features/verify.js`), so with `--skip-connect` nothing consumed it — yet the fetch still ran, and its failure was recorded as an error-level finding that failed the whole run. A source no enabled check used was breaking CI.

**2. The fetch was unauthenticated against a private repo.** `rp-connect-docs` and `docs` are both private, and `raw.githubusercontent` answers unauthenticated requests for private content with **404, not 401** — so a permissions problem was indistinguishable from a missing file.

## Changes

- `--skip-connect` now also skips the `antora.yml` fetch
- `fetchText` sends a bearer token when one is resolvable through the existing `cli-utils/github-token` chain (`GIT_CREDENTIALS` → `REDPANDA_GITHUB_TOKEN` → `ACTIONS_BOT_TOKEN` → `GITHUB_TOKEN` → …)
- a tokenless 404 now names the likely cause instead of leaving the reader to guess

The unnamed-source path still throws rather than returning `undefined`, so a missing `--registry` keeps failing hard.

## Verification

Run against the real registry and disable page taken from the docs branch:

| scenario | result |
|---|---|
| before, `--skip-connect`, no token | exit 1 — the exact CI error |
| after, same inputs | **exit 0**, no fetch failure |
| after, connect enabled, no token | same 404, now with the token hint |
| after, connect enabled, with token | private `antora.yml` resolves, zero 404s |

Confirmed independently that `Authorization: Bearer` works on `raw.githubusercontent` for a private repo (404 → 200).

## Note

This unblocks the `validate` check on docs#1895/#1896 with no change needed in those PRs, since their workflow already passes `--skip-connect`. Once released, that check should pass without a token. A token only becomes necessary if the connect comparison is re-enabled by dropping `--skip-connect`.

### DIFF
--- a/bin/doc-tools.js
+++ b/bin/doc-tools.js
@@ -2165,17 +2165,28 @@ validation
     const yamlLib = require('js-yaml')
 
     const RAW = 'https://raw.githubusercontent.com'
+    // Some sources live in private repos (docs, rp-connect-docs), which
+    // raw.githubusercontent serves as 404 rather than 401 when unauthenticated
+    // — indistinguishable from a genuinely missing file. Send a token when one
+    // is available so those fetches resolve; public sources are unaffected.
+    const { getGitHubToken } = require('../cli-utils/github-token')
+    const ghToken = getGitHubToken()
     // Sources that fail to load are reported as error-level findings rather
     // than silently skipped, so CI cannot pass on a check that never ran.
     const failedSources = []
     async function fetchText (url, sourceName) {
-      const resp = await fetch(url)
+      const resp = await fetch(url, ghToken ? { headers: { Authorization: `Bearer ${ghToken}` } } : undefined)
       if (!resp.ok) {
+        // A 404 with no token is the signature of an unauthenticated private
+        // repo, so say so instead of leaving it to look like a missing file.
+        const hint = resp.status === 404 && !ghToken
+          ? ' If this source is in a private repository, set GITHUB_TOKEN (or REDPANDA_GITHUB_TOKEN / ACTIONS_BOT_TOKEN) so the fetch can authenticate.'
+          : ''
         if (sourceName) {
-          failedSources.push({ level: 'error', check: 'fetch', message: `Could not load ${sourceName} (${url}): ${resp.status} ${resp.statusText}. The related checks did not run.` })
+          failedSources.push({ level: 'error', check: 'fetch', message: `Could not load ${sourceName} (${url}): ${resp.status} ${resp.statusText}. The related checks did not run.${hint}` })
           return undefined
         }
-        throw new Error(`Failed to fetch ${url}: ${resp.status} ${resp.statusText}`)
+        throw new Error(`Failed to fetch ${url}: ${resp.status} ${resp.statusText}.${hint}`)
       }
       return resp.text()
     }
@@ -2193,9 +2204,14 @@ validation
       const disablePage = options.disablePage
         ? readLocal(options.disablePage)
         : await fetchText(`${RAW}/redpanda-data/docs/${options.docsRef}/modules/get-started/pages/licensing/disable-enterprise-features.adoc`, 'disable-enterprise-features.adoc')
-      const antoraYaml = options.antora
-        ? readLocal(options.antora)
-        : await fetchText(`${RAW}/redpanda-data/rp-connect-docs/main/antora.yml`, 'rp-connect-docs antora.yml')
+      // Only checkConnect consumes this, so --skip-connect has to skip it too.
+      // Fetching it anyway meant --skip-connect still reached a private repo
+      // and failed the run on a source none of the enabled checks used.
+      const antoraYaml = options.skipConnect
+        ? undefined
+        : options.antora
+          ? readLocal(options.antora)
+          : await fetchText(`${RAW}/redpanda-data/rp-connect-docs/main/antora.yml`, 'rp-connect-docs antora.yml')
       const antoraEnterpriseComponents = antoraYaml
         ? extractAntoraEnterpriseComponents(yamlLib.load(antoraYaml))
         : undefined


micheleRP (commented):
## Review: honor `--skip-connect` and authenticate private fetches

Correct fix, accurately diagnosed. I verified both root causes independently rather than taking the description on trust, and found no code defects. Holding off on approval only because of the release step below — the code itself is good to go.

### Verified

| Claim | Result |
|---|---|
| Only `checkConnect` consumes `antoraEnterpriseComponents` | Confirmed — `verify.js:226`, gated on `sources.infoCsv` at `:224`, so under `--skip-connect` the fetched value was genuinely dead weight |
| `Authorization: Bearer` resolves private raw content | Confirmed live: 404 unauthenticated → 200 with the header (the `token` scheme also works) |
| `docs` and `rp-connect-docs` are private | Confirmed both; `redpanda` and `connect` are public, so the core-header and info.csv fetches need no auth |
| The check passes with no token after this | Confirmed — the workflow passes `--registry` and `--disable-page` locally plus `--skip-connect`, leaving only the two public core-header fetches |

### Needs doing before this helps CI

**1. Bump `package.json` to 5.11.1.** `publish-to-npm.yaml` publishes on push to `main` only when the version is new, and this branch leaves the version at 5.10.0 while `main` is already at the published 5.11.0. So merging as-is publishes nothing and the fix never reaches the workflow it exists to fix. docs#1895's lockfile then needs a matching re-bump — it currently resolves exactly 5.10.0, which `npm ci` installs verbatim.

Also worth rebasing: the branch is 3 commits behind `main`.

### Suggestions

**2. The hint only fires when no token is present.** `hint` is gated on `!ghToken`, so whenever a token exists but cannot read the repo you get a bare 404 again — the same ambiguity this PR set out to remove, relocated to the other branch.

This is now the live path rather than a hypothetical: rp-connect-docs#485 has merged, which the workflow's TODO names as the trigger for dropping `--skip-connect`. The moment that happens, the fetch crosses into a private sibling repo, where the default Actions `GITHUB_TOKEN` is scoped to its own repository and will 404 with a token present. Suggest branching the message: no token → set one; token present → it may not grant access to that repo.

**3. The hint omits `GIT_CREDENTIALS`.** It outranks every variable the message does name, and it is the one actually set in these workflows — docs-site#203 exports exactly `GIT_CREDENTIALS: https://git:$TOKEN@github.com` for cross-repo access. Worth naming first.

**4. `--skip-connect`'s help text now understates the flag.** It reads "Skip the connect info.csv comparison", but the flag now also skips the `rp-connect-docs` antora.yml fetch, and silently overrides an explicitly passed `--antora <path>`. Harmless today, since `checkConnect` cannot run without info.csv, but surprising.

**5. Optional — testability.** `verify.js` has a unit-test suite; this precedence chain sits inline in the Commander action, the one layer nothing covers, and it is exactly what regressed.

### On the follow-up direction

Now that #485 has removed the `enterprise-components` list (I confirmed the key is gone from `antora.yml` on main), dropping `--skip-connect` would fetch a private repo purely for `extractAntoraEnterpriseComponents` to return `undefined` and `checkConnect` to return immediately at `verify.js:130`. The comparison would cost a cross-repo credential and verify nothing. Retiring the antora.yml source looks like the better follow-up than dropping the flag. One piece of luck: the parser returns `undefined` rather than `[]`, so there is no flood of false "missing from the list" errors.

### Related, out of scope

`cli-utils/self-managed-docs-branch.js:13` has the same root cause unfixed — `fetchRemoteAntoraVersion()` fetches the now-private `docs` antora.yml via bare `https.get` with no token support at all. I ran it: it rejects. Its caller at `bin/doc-tools.js:1929` hard-exits 1, so `doc-tools generate crd` fails at that step when run inside the docs repo. Worth a separate issue.

### What works well

- The two root causes really are independent, and the 404-not-401 behavior on private raw content is the non-obvious one. It reproduces exactly as described.
- Token goes in a header rather than the URL, so it cannot leak into the error messages that print the URL.
- The unnamed-source throw path is deliberately preserved, so a bad `--registry` still fails hard instead of silently yielding `undefined`.
- Comments explain why the gating and the hint exist, not what the code does.

### CodeRabbit

The `resp.body?.cancel()` suggestion is correct about undici in general but low value here: this run makes about six fetches and both terminal paths call `process.exit()`, so there is no accumulation and no exit hang. Fine to skip.

micheleRP:
### Follow-up: three findings from running the command

I went back over this by actually executing `validate enterprise-features` against the real registry under each token state, rather than reading the diff. The fix does what it claims — worth stating plainly:

| Scenario | Result |
|---|---|
| this branch + `--skip-connect`, no token | `exit 0`, "Registry is in sync with all checked sources" |
| `main` + `--skip-connect`, no token | `exit 1`, the exact CI error verbatim |
| this branch, connect enabled, no token | `exit 1`, 404 with the new hint |
| this branch, connect enabled, with token | `exit 0`, all four fetches 200, private antora.yml resolves |

Three things surfaced that reading the diff did not.

---

**1. The private-repo hint can point at the wrong cause.**

While testing during today's GitHub outage, `configuration.h` — from the **public** `redpanda` repo — returned a transient 404 and failed the run. It reproduced for several minutes before clearing on its own.

The hint fires on `resp.status === 404 && !ghToken`, which has no knowledge of whether the source is actually private. Three of the four remote sources here are public (both core headers, connect info.csv), so the same transient blip without a token appends:

> If this source is in a private repository, set GITHUB_TOKEN (or REDPANDA_GITHUB_TOKEN / ACTIONS_BOT_TOKEN) so the fetch can authenticate.

...to a file that needs no authentication, sending the reader after a credential problem that does not exist. Since the private sources are known statically (`docs`, `rp-connect-docs`), gating the hint on those — or softening it to name both possibilities — would keep it accurate.

Worth knowing if you re-run these tests today: I nearly concluded "the token breaks public fetches" off the back of that 404. It reproduced twice, and only a curl comparison plus a traced re-run showed it was the outage rather than the change.

**2. No retry on transient fetch failures.**

Any single blip on any of the four remote sources becomes an error-level finding and fails the whole run, which is exactly what I hit above. #259 added `--retry 3 --retry-delay 2` to curl in this same repo for precisely this class of problem, so there is already a house precedent to follow. Worth the same treatment here, especially given the check runs on a weekly cron where nobody is watching live.

**3. The connect comparison is silently vacuous — measured, not inferred.**

rp-connect-docs#485 merged and removed the `enterprise-components` list. With connect enabled and a valid token, the antora.yml fetch now succeeds (14 KB), `extractAntoraEnterpriseComponents` returns `undefined`, and `checkConnect` compares **36 enterprise plugins from info.csv against nothing**:

```
total findings:      2
connect-list checks: 0     <-- the comparison ran and reported nothing
drift:               false
checks represented:  core-property
```

No "skipped" line either, because info.csv *was* provided — `runChecks` only emits that when info.csv is absent. So the run looks clean while having verified nothing about connect.

The mildly awkward consequence is that `--skip-connect` is currently the more honest mode: it at least emits `INFO [connect-list] Skipped: no connect info.csv provided.` This is really an argument for retiring the antora.yml source rather than acting on the workflow's TODO to drop the flag — dropping it would buy a cross-repo credential requirement in exchange for a check that reports nothing.

None of this blocks the fix in this PR, which is sound. Flagging 1 and 2 as small hardening opportunities while the file is open, and 3 as context for whoever picks up the connect follow-up.

micheleRP:
@JakeSCahill one small thing and I'll approve this — the code itself needs no changes, I verified it works by running it.

It needs a version bump to publish. `publish-to-npm.yaml` fires on push to `main` but only publishes when the version is new, and this branch is still at 5.10.0 while `main` is already at 5.11.0 (published). So as it stands, merging this is a no-op for consumers and docs#1895's `validate` check stays red — which is the thing this PR is meant to fix.

Cleanest order, since `main` has moved the version out from under the branch:

1. Rebase onto `main` (3 commits behind, and it picks up 5.11.0 — otherwise the version line conflicts on merge)
2. Bump `package.json` to **5.11.1** — patch fits, it's a `fix`

Then after it publishes, docs#1895 needs its lockfile moved to 5.11.1 as well. It currently resolves exactly 5.10.0 and CI runs `npm ci`, so bumping the caret range alone won't pick the fix up.

Ping me once that's pushed and I'll approve straight away. The three findings in my previous comment are all optional hardening — nothing there needs to hold this up.

micheleRP (approved):
Approving the code — to be clear about what this does and doesn't cover, since my earlier review said I was holding off.

The change itself is correct and I verified it by running the command rather than only reading the diff: with `--skip-connect` and no token it exits 0, and `main` under the same inputs exits 1 with the exact CI error. Both root causes hold up, including the non-obvious one where raw.githubusercontent answers private content with 404 rather than 401.

Two things this approval is not gating on, both already noted above:

- The rebase and version bump to 5.11.1. Still needed before merging, or the merge publishes nothing and docs#1895 stays red — but that's release mechanics rather than a code problem, and it's yours to sequence.
- The three findings in my earlier comment (hint accuracy, no fetch retry, the now-vacuous connect comparison). All optional hardening, none blocking.

Good fix. The diagnosis of two independent causes behind one error message was the hard part here, and the comments explaining why each guard exists will save the next person the same investigation.

JakeSCahill:
@micheleRP All three of your findings from the follow-up comment are now addressed in `bce0daa`, plus the earlier asks (5.11.1 bump, `GIT_CREDENTIALS`-first hint, help text) landed in `538dda4`:

1. **Hint accuracy** — credential hints are now gated on the statically-known private repos (`docs`, `rp-connect-docs`). A 404 on a public source (core headers, info.csv) reports as a plain 404 — no more credential wild-goose chase during a GitHub blip.
2. **Transient retry** — network errors, 429s, and 5xx now retry 3× with a 2s delay (the #259 curl `--retry` precedent), so a single blip no longer fails the weekly cron run.
3. **The vacuous connect check** — went further than context-noting it: the antora.yml source is retired outright. `checkConnect` now validates the registry's `connect-plugin` entries against info.csv and *reports what it verified* (`info.csv lists 36 enterprise plugin(s); N pinned by registry connect-plugin entries…`), so a shallow pass is visible instead of silent. `--antora` is gone; `--skip-connect` now only skips the info.csv fetch.

Nice side effect: with the private source retired, the **full connect check runs tokenless** (verified locally, exit 0 with all six token vars unset) — so docs#1895 can drop `--skip-connect` after this publishes, with no cross-repo credential and a check that actually verifies something.

Verified on the branch: 59 suites / 1352 tests green; full behavioral matrix re-run (tokenless skip-connect, rejected-token retry, full connect check with and without token — all exit 0); `npm pack` of this branch installed into docs#1895's checkout runs its validate workflow command clean.

Re-requesting your review since the new commits invalidated the approval.

JakeSCahill:
One more hardening on top: `--connect-ref` now defaults to `latest` (resolved to the newest connect release tag via the GitHub API) instead of `main` — the registry documents released state, so unreleased renames/additions on main should not skew the check. The resolved ref is reported in the findings (`info.csv@v4.105.0 lists 36 enterprise plugin(s)...`), a failed resolution is an error finding rather than a silent fallback, and explicit refs (including `main`) still work. Core headers stay on `dev` deliberately — that check exists for early warning of new license gates. 59 suites / 1357 tests green.

JakeSCahill:
Follow-up per Jake: if the GitHub API can't resolve `releases/latest` (outage, rate limit, missing `tag_name`), the connect ref now falls back to the `latest-connect-version` asciidoc attribute in rp-connect-docs `antora.yml` — the newest version the docs actually document, i.e. the state the registry describes (currently `4.103.1`). The fallback is announced with a warning; only if both routes fail does the connect check skip with an error finding. Never a silent fallback to `main`. 59 suites / 1359 tests green.

